PR #7, opened 2021-01-20
Fix to PMS_Init and add new example ADVANCED_WITH_OTA

init99:
Else was not attached to the correct if statement. 

Open question whether the return value should be tested for <0 vs <=0 ?


init99:
Adding in a new example that includes OTA, F/C temp setting, temp only capability, tagged logging, update rate backoff, a way to reset the wifi using the serial connection, and presentation of the PM2.5 and CO2 labels as sent from Air Gradient.

PR #8, opened 2021-05-29
correct math.h header and specify wificlient as per new http library requirement.

markasoftware:
The library and example do not work for me without these fixes.


mudit31:
I have verified that this change works, and is needed.

Issue #9, opened 2021-09-11
Missing return in periodicFetchData ?


chrissearle:
First up - the code compiles fine in the usual arduino IDE.

But - in platformio - where it builds with certain warnings enabled - in this case -Werror=return-type - we get:

```
AirGradient.cpp: In member function 'TMP_RH AirGradient::periodicFetchData()':
AirGradient.cpp:306:16: error: control reaches end of non-void function [-Werror=return-type]
  306 |     returnError(error);
```

I can fix it locally by just changing line 306 to:

```
return returnError(error);
```

I guess that the arduino IDE doesn't enable that warning.

But - it does look like a return on that line would be a good idea?

Not entirely sure - but thought I'd mention it.

Nightreaver:
I can confirm this issue - it should be fixed for platformio if it doesnt bother arduino IDE anyways

airgradienthq:
Thank you. It has been merged and is part of the latest release (1.4.0)

PR #12, opened 2021-10-15
Added return to compile with platformIO

StrongholdIII:
Solves https://github.com/airgradienthq/arduino/issues/9


oparaskos:
I also ran into this

@airgradienthq @boomacoh Can this be merged?

PR #15, opened 2021-11-03
add influx 1 example

hez:
This is similar to `examples/C02_PM_SHT_OLED_WIFI` but it pushes to an InfluxDB instead of Prometheus.  Thought it might be helpful for others out there!

I made a few quality of life improvements line 40-41: defined the refresh rate for screen and push (don't want to push every 3seconds to the db).

Cheers

### DIFF
--- a/examples/C02_PM_SHT_OLED_WIFI_INFLUX/C02_PM_SHT_OLED_WIFI_INFLUX.ino
+++ b/examples/C02_PM_SHT_OLED_WIFI_INFLUX/C02_PM_SHT_OLED_WIFI_INFLUX.ino
@@ -38,14 +38,16 @@ AirGradient ag = AirGradient();
 SSD1306Wire display(0x3c, SDA, SCL);
 
 #define DISPLAY_DELAY 3000
-#define INFLUX_SEND_PERIOD 5 * 60 * 1000
-static unsigned long lastSampleTime = 0 - INFLUX_SEND_PERIOD;  // initialize such that a reading is due the first time through loop()
+// Number of times to run full loop before resending to influx
+#define INFLUX_SEND_PERIOD 37
+static unsigned long current_loop = INFLUX_SEND_PERIOD;  // initialize such that a reading is due the first time through loop()
 
 
 // set sensors that you do not use to false
-boolean hasPM=true;
-boolean hasCO2=true;
-boolean hasSHT=true;
+const boolean hasPM=true;
+const boolean hasCO2=true;
+const boolean hasSHT=true;
+const float shtTempOffset=-2.4;
 
 // set to true if you want to connect to wifi. The display will show values only when the sensor has wifi connection
 boolean connectWIFI=true;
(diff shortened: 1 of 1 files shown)


examples/C02_PM_SHT_OLED_WIFI_INFLUX/C02_PM_SHT_OLED_WIFI_INFLUX.ino
```diff
@@ -0,0 +1,192 @@
...
+  if (connectWIFI) connectToWifi();
+  client.setConnectionParamsV1(INFLUXDB_URL, INFLUXDB_DB_NAME, INFLUXDB_USER, INFLUXDB_PASSWORD);
+  delay(2000);
+}
+
+void loop(){
+  unsigned long now = millis();
+  const boolean shouldPushToInflux = now - lastSampleTime >= INFLUX_SEND_PERIOD;
+  if (now - lastSampleTime >= INFLUX_SEND_PERIOD)
+    lastSampleTime += INFLUX_SEND_PERIOD;
```
McJoppy, line 84: Hi - looking at pushing metrics to influxDB and came across this.

FYI millis() will overflow to 0 after around 50 days - once this happens it looks like this will fall apart one `now` resets as `now - lastSampleTime` would result in a negative value.

I may be wrong :)

FYI https://www.arduino.cc/reference/en/language/functions/time/millis/

hez: @McJoppy That is a good point, I hadn't considered that.  Open to suggestions here, might just make it a counter and change the setting to a multiple of the main loop sleep.

McJoppy: I'm about to try this logic

**Setup**

- test sendPeriod not 0
- init 'lastDividedWhole' as unsigned long

**Loop**

- set now to millis()
- divide now by sendPeriod and assign to unsigned long sendTest
- if sendTest != lastDividedWhole then period must have passed

McJoppy: My take on sending to InfluxDB can be found at https://gist.github.com/McJoppy/aab9e1f9e2e43fdea61eabe729122635

I tend to hack stuff together and over complicate stuff. I've tried to keep the original functionality there but disabled by default.

Also beware of line 50. I'm not a fan of the default font so have added my own Roboto which provides better info for me.

My remote server supports v1 so currently no support for v2 too.

Hope this is useful :)

hez: I pushed up a commit removing the use of `millis`. https://github.com/airgradienthq/arduino/pull/15/commits/001e15bcbf9e560256fbc27fdb99c44c3e6eaf00

danepowell:
Thanks, I ended up writing my own InfluxDB adaptation before running across this PR, it would have been nice to have out of the box.

In my implementation, I send a single point with multiple fields:
```
Point sensor("airgradient");
void loop() {
  sensor.clearFields();
  sensor.addField("pm2", PM2);
  sensor.addField("co2", CO2);
  client.writePoint(sensor);
}
```

I'm relatively new to InfluxDB, is there a reason to prefer using separate points as you did? Not sure of best practices. Regardless, it seems like it would be good to batch calls.

McJoppy:
Yep - the, one point is best :)

FYI the code in the example also sends a single point... 

```
// matching constant type based off info from AirGradient.h
void sendInfluxData(const int pm2, const int co2, const float temp, const int humidity) {
  influxSensor.clearFields();
  influxSensor.addField("pm2.5", pm2);
  influxSensor.addField("co2", co2);
  influxSensor.addField("temperature", temp);
  influxSensor.addField("humidity", humidity);
  
  // Print what are we exactly writing
  Serial.print("Writing: ");
  Serial.println(influxClient.pointToLineProtocol(influxSensor));

  // Write point
  if (!influxClient.writePoint(influxSensor)) {
    Serial.print("InfluxDB write failed: ");
    Serial.println(influxClient.getLastErrorMessage());
  }
}
```

airgradienthq:
I cannot merge this because we have now a complete rewrite of the library published. I also would like to keep the examples here lean and focused and not have too many options that could confuse people new to it.

PR #20, opened 2022-02-05
Major refactor of the library to support easily new sensors

Belphemur:
Hi,

As discussed on the forum, I've rework the library from scratch to be more dynamic and adapt to different sensors.

The main idea (as shown in examples) is to rely on `MetricGatherer` to do all the heavy lifting.

# Concept
Each sensor has its own way to work. The `MetricGatherer` will ask all of them to setup and will gather the data at fixed interval.

## Folders
* `src/Sensors` contains all the different supported sensors.
* `src/Prometheus` to provide a metrics endpoint that prometheus can easily scrape.
* `src/AQI` provide AQI calculation by first averaging in 24h the PM 2.5 (as specified in the spec of AQI) then applying the AQI formula on that average.
* `src/NTP` provide a basic NTP client, used to give a bootime (useful to know for how long AirGradient has been running)
* `src/Metrics` where lies the `MetricsGatherer`.

# Usage
```cpp
auto metrics = std::make_shared<MetricGatherer>();
void setup() {
  Serial.begin(9600);

	// Here we add the different sensors we have on the board.
   // This will check that we don't add twice a sensor that provide the same data (like 2 CO2 sensor)
    metrics->addSensor(std::make_unique<PMSXSensor>())
            .addSensor(std::make_unique<SHTXSensor>())
            .addSensor(std::make_unique<SensairS8Sensor>());

// Advanced usage where the dev don't want all the measurement that a sensor can do (like with the TVOC that can provide also CO2). Here we set it up to provide everything BUT CO2.
        metrics->addSensor(std::make_unique<SGP30Sensor>(), Measurement::CO2)

  // This take care of intializing the different sensor
  // The gatherer will take care of pooling the data every 3s (configurable in it's constructor).
   metrics->begin();
}
void loop() {
 //Contain the different measurement provided by the sensor
 auto data = metrics->getData();
//This contains the different type of sensor that have been configured
 auto sensorType = metrics->getSensorTypes();

//Here we check that we have a Particle sensor configured in the MetricGather
//This is to avoid getting unused data
  if (!(sensorType & SensorType::Particle)) {
                Serial.printf("PM2.5: %f", data.PARTICLE_DATA.PM_2_5);
    }
}
```
# Todo after merge
- [ ] Bump the version of the library to 2.0.0 since it's a major breaking change


airgradienthq:
This has been addressed in version 3 of the library that has now a complete modular concept and integrated most of the necessary libraries.

PR #23, opened 2022-11-05
Add flipDisplay configuration setting

kerrickstaley:
Add a `flipDisplay` configuration setting. When set to true (the default), the display will be flipped vertically.

If you are using the [official case](https://www.airgradient.com/open-airgradient/instructions/the-airgradient-diy-sensor-enclosure/), you want to set `flipDisplay = false`.


airgradienthq:
This should be addressed in v3 of the library.

PR #26, opened 2023-02-22
Update AirGradient.cpp to use proper documented PDU and function code…

ken830:
… for CO2 reading from S8 Sensor

New PDU is now 8 bytes long. Added responseSize = 7 to continue to support a 7-byte response


airgradienthq:
Closing this as we have now completely rewritten the firmware.

PR #28, opened 2023-04-14
Use modbus library to communicate with CO2 sensor

Lalufu:
The SenseAir S8 used in AirGradient is accessed via the modbus protocol, over an RS232 serial bus.
The current implementation has dealt with this through a minimal modbus layer, which has led to a number of problems in the past (problems with timing, incorrect readings).

This patch implements the CO2 sensor access through a modbus library, which implements the entire low level communication and provides proper checksum testing and other error handling.

The changes are in the `AirGradient::getCO2_Raw` method, and the external behaviour of the method is unchanged.

The modbus library chosen is emelianov/modbus-esp8266, which needs to be added to the dependency list in Arduino.

This change increases the size of the sketch by roughly 8kB on an ESP8266 board.

This change currently gates the new implementation behind a feature flag (`AIRGRADIENT_MODBUS`) for easier testing. This macro needs to be defined in the compile stage.

I have run a sensor with this change for a week, and have encountered no further spurious CO2 readings, while before there was roughly one per day.

### DIFF
--- a/AirGradient.cpp
+++ b/AirGradient.cpp
@@ -714,8 +714,13 @@ void AirGradient::CO2_Init(int rx_pin,int tx_pin,int baudRate){
     }
   _SoftSerial_CO2 = new SoftwareSerial(rx_pin,tx_pin);
   _SoftSerial_CO2->begin(baudRate);
+#ifdef AIRGRADIENT_MODBUS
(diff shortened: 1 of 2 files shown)


airgradienthq:
Closing this as we have vastly improved communication with CO2 module in the meantime.

PR #36, opened 2023-12-02
Fix compilation warning

jschneier:
In the Arduino IDE there are compilation errors like:

```
AirGradient_Air_Quality_Sensor/AirGradient.cpp:684:14: warning: converting to non-pointer type 'float' from NULL [-Wconversion-null]
  684 |   result.t = NULL;
      |              ^~~~
AirGradient_Air_Quality_Sensor/AirGradient.cpp:685:15: warning: converting to non-pointer type 'int' from NULL [-Wconversion-null]
  685 |   result.rh = NULL;
```

Just setting to `-128` since they are obviously wrong.


airgradienthq:
Closing this as we have now much better firmware in place.

PR #37, opened 2023-12-10
Fix: Timing between loop actions

Manezki:
Use the current time for timestamping the previous action time.

This prevents bursting behaviour in case of `currentMillis - previousAction > 2*actionInterval`.
Additionally the change should make the logic slightly more readable.

## Testing

* Tested on ONE_V9.ino, which maintains measurement interval close to `actionInterval`.

## Notes

While `previousAction = millis();` would be more readable and precise, that caused AirGradient One to sample TVOC about every 2000millis compared to 1000millis that the current change does.


airgradienthq:
I believe this has been addressed with version 3 of the library.

PR #45, opened 2024-02-03
Handle humidity measurement as a float

mtlynch:
The `getHumidity` function of the arduino-sht library returns a float

https://github.com/Sensirion/arduino-sht/blob/5e89ae8b9b7eb16675e02d73c417dd79f64b47ae/SHTSensor.h#L148

When AirGradient assigns the return value of `getHumidity` to an `int`, it truncates the decimal portion of the value and reduces precision of the reading. This patch adjusts the type of the `hum` field to a `float` so that we capture the full precision from the SHT sensor.

### DIFF
--- a/examples/ONE_V9/ONE_V9.ino
+++ b/examples/ONE_V9/ONE_V9.ino
@@ -128,7 +128,7 @@ int pm10 = -1;
 const int tempHumInterval = 5000;
 unsigned long previousTempHum = 0;
 float temp;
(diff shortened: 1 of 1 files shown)


airgradienthq:
I cannot merge this because we have now a complete rewrite of the library published and I believe this issue should be resolved. If not please make a new PR based on v3+ of the library. Many thanks!

PR #56, opened 2024-02-24
Fix path capitalization

dechamps:
This fixes build breakage on case-sensitive filesystems (e.g. Linux) with errors like:

```
src/Display/Display.h:4:10: fatal error: ../main/BoardDef.h: No such file or directory
```

### DIFF
--- a/src/Display/Display.h
+++ b/src/Display/Display.h
@@ -1,7 +1,7 @@
 #ifndef _AIR_GRADIENT_OLED_H_
 #define _AIR_GRADIENT_OLED_H_
 
-#include "../main/BoardDef.h"
+#include "../Main/BoardDef.h"
 #include <Arduino.h>
 #include <Wire.h>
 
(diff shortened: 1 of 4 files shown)


pnt325:
Hi @dechamps 
Thank for check and report this issue. Can you provide more info about the build environment that make issue. When we modify the folder as upcase and build project on windows without any issue that make missing update the file path

Thank,
Phat

dechamps:
> Can you provide more info about the build environment that make issue.

It's only an issue when building from a case-sensitive filesystem, such as those used on Linux (e.g. ext4), and (presumably) Mac.

> When we modify the folder as upcase and build project on windows without any issue that make missing update the file path

Windows is case-insensitive so yes, that would work on Windows just fine, but not on other systems. This is not surprising.

The CI checks added in #58 will catch any such issues in the future, as the CI runs on Linux.

PR #70, opened 2024-02-29
Fix exclude list in GA check

dechamps:
This was not updated in a59d5a1bb8510f85a306daf14a879489cc364ad9. This fixes GitHub Actions breakage.


dechamps:
(Hmm, I can see on this PR that the GitHub Actions workflow is not working on PRs, though it appears to work on pushes, i.e. when running directly off a branch. I think this has to do with subtle differences in `actions/checkout@v4` behavior when running off a PR vs. running off a branch; see actions/checkout#426. That seems to confuse the subsequent `arduino-cli lib install --git-url .` command. I'm looking into it and will sent a separate PR to fix it. In the mean time, [this run](https://github.com/dechamps/airgradient/actions/runs/8103557202) can be used as proof that this PR fixes the original breakage.)

dechamps:
(The unrelated PR workflow failures are addressed by #71.)

PR #71, opened 2024-02-29
Fix check workflow failing on pull requests

dechamps:
This fixes the following check GitHub Actions workflow failure that would otherwise occur on pull requests (but not on pushes/branches):

```
Error installing Git Library: Library install failed: object not found
```


dechamps:
(The two isolated failures are expected and are separately fixed by #70.)

PR #78, opened 2024-03-08
[ONE/Prometheus] Use full unit in temperature metric

gouthamve:
Please see: https://prometheus.io/docs/practices/naming/#base-units

Also, thanks a lot for this support, I had my own exporter that I can now delete :)


gouthamve:
cc @dechamps

dechamps:
Good catch! I tried to follow all the best practices but it looks like I missed that one :(

PR #90, opened 2024-03-29
reset watchdog after sending mqtt message

lschloetterer:
When disabling the serverSchedule task the ESP always resets itself after ~15 minutes.
With the additional call to resetWatchdog() in the mqtt task everything works fine.

### DIFF
--- a/examples/ONE/ONE.ino
+++ b/examples/ONE/ONE.ino
@@ -1235,6 +1235,7 @@ static void createMqttTask(void) {
             if (agMqtt.publish(topic.c_str(), syncData.c_str(),
                                syncData.length())) {
               Serial.println("MQTT sync success");
+              resetWatchdog();
             } else {
               Serial.println("MQTT sync failure");
             }


pnt325:
Hi @lschloetterer 

Thank for your PR. Currently we has add the local reset external watchdog timer if device could not connect to server. So we don't need to add reset after MQTT send success.

Please update to newest version: [3.0.9](https://github.com/airgradienthq/arduino/releases/tag/3.0.9)

lschloetterer:
Thank you for your reply.
I think you are talking about the wdgFeedSchedule? If I understand the code correctly, it runs only when the offlineMode flag is true, which is only set when no WiFi connection is available. To use mqtt however, a local network connection is required.

I'm not sure what the purpose of the watchdog is, maybe it would make sense to run the reset schedule always, not just in offlineMode?

pnt325:
You're it's only support in offline mode(Without WiFi configuration or can not connnect).
External watchdog timer is use for auto reset device if software has issue and stop work. Currently we're update support local config device over HTTP.

examples/ONE/ONE.ino
```diff
@@ -1235,6 +1235,7 @@ static void createMqttTask(void) {
             if (agMqtt.publish(topic.c_str(), syncData.c_str(),
                                syncData.length())) {
               Serial.println("MQTT sync success");
+              resetWatchdog();
```
nick-4711, line 1238: I propose to centralize the watchdog reset feature and feed in information such as e.g. "MQTT publishing worked" and then let it decide on criteria such as if e.g. posting to the AG backend is enabled if it should reset or not. @pnt325 can you look into this?

pnt325: I'll check it. 

On newest version I have callback direct to `ag->watchdog.reset()` for reset.

airgradienthq: @pnt325 What is the status here. Has this been addressed and can I close this PR?

pnt325: > @pnt325 What is the status here. Has this been addressed and can I close this PR?

The watchdog reset by post data to AG success or offline mode. It's not reset by MQTT publish. We don't have any MQTT connect to AG. The MQTT connection by user. I think we can close it

airgradienthq:
Has been addressed in the meantime with fixed interval of wd reset.

PR #91, opened 2024-03-29
allow setting hue of LedBar

lschloetterer:
This exposes the hue color functionality from the NeoPixel library.

I use it to set the CO2 color gradually like this:

```cpp
void setRGBledCO2color(int co2Value) {
  if (co2Value <= 0) return;
  co2Value = min(co2Value, 3000); // don't calculate hue beyond 3000 to not return green again
  uint16_t hue = map(co2Value, 500, 1800, UINT16_MAX/3 /*green*/, 0 /*red*/);
  ag.ledBar.setHue(hue);
}
```

### DIFF
--- a/src/Main/LedBar.cpp
+++ b/src/Main/LedBar.cpp
@@ -111,6 +111,17 @@ void LedBar::setColor(uint8_t red, uint8_t green, uint8_t blue) {
 }
 
 /**
+ * @brief Sets the hue of all LEDs
+ *
+ * @param hue Hue of the color as 16-bit number
+ */
+void LedBar::setHue(uint16_t hue) {
+  uint32_t rgbcolor = pixel()->ColorHSV(hue);
+  pixel()->fill(rgbcolor);
+  pixel()->show();
+}
+
+/**
  * @brief Call to turn LED on/off base on the setting color
  *
  */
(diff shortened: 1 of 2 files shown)


src/Main/LedBar.cpp
```diff
@@ -110,6 +110,17 @@ void LedBar::setColor(uint8_t red, uint8_t green, uint8_t blue) {
   }
 }
 
+/**
+ * @brief Sets the hue of all LEDs
+ *
+ * @param hue Hue of the color as 16-bit number
+ */
+void LedBar::setHue(uint16_t hue) {
```
nick-4711, line 118: Should we also support to set saturation and brightness, or is this not of interest?

src/Main/LedBar.cpp
```diff
@@ -110,6 +110,17 @@ void LedBar::setColor(uint8_t red, uint8_t green, uint8_t blue) {
   }
 }
 
+/**
+ * @brief Sets the hue of all LEDs
```
nick-4711, line 114: I haven't worked with the HSV color model yet, but I would expect that we want to set this on a per LED base (and not for all).

nick-4711: @lschloetterer

lschloetterer: You're right, having the possibility to individually set the LEDs would be more dynamic.

However I see now that the code for the LED CO2 handler isn't in the arduino sketch anymore, but moved into the library in the AgStateMachine class. So having a custom handler isn't possible at the moment without modifying the library code anyway.

I'm not sure if there is a good solution to provide the possibility to have a custom LED handler function, I'd have to look into this when I have some more time

airgradienthq: @lschloetterer I am closing this PR for now. Please feel free to open a new PR at a later stage which is compatible with the current firmware.

PR #92, opened 2024-04-07
Use capped 'val' value in displayShowDashboard()

Flowdalic:
Fixes: e47096feac33 ("Limit show CO2 index within 4 character number (max = 9999)")


pnt325:
Hi  @Flowdalic 
Thank for your PR. Currently we have developing new firmware to support both ESP32C3 indoor and outdoor. We're also remove the example `ONE.ino` and `OpenAir.ino`. You're issue also fixed [here](https://github.com/airgradienthq/arduino/blob/develop/src/AgOledDisplay.cpp)

PR #96, opened 2024-04-29
Report PM data to metrics API for AirGradient ONE

danielmoore:
I noticed that PM 0.1, 2.5, and 1.0 data on my AirGradient ONE was being reported by `/measures/current` but not `/metrcis`. This MR brings those two endpoints better into parity.

I also noticed a comma went missing causing Prometheus to no longer be able to parse the metrics. Fixed that, too.

As a side note, I notice that the logic for handling a measurement seems to be duplicated between `AgValue.cpp` and `OpenMetrics.cpp`. I suspect we'll keep seeing this kind of disparity between these endpoints until that logic is unified.


pnt325:
Hi @danielmoore 
Thank for your PR, and also the notes, we will check the code and also our side.

Phat

PR #102, opened 2024-05-01
fix led bar button test not work on power up

pnt325:


src/AgStateMachine.cpp
```diff
@@ -302,20 +302,24 @@ void StateMachine::co2Calibration(void) {
...
-          break;
-        }
-      }
-    } else {
+    ledBarPowerUpTest();
+  }
+}
+
+void StateMachine::ledBarPowerUpTest(void) {
+  if (config.getCountry() == "TH") {
```
nick-4711, line 310: I think we shouldn't have a different behavior based on the country setting

pnt325: There are 3 inputs parameter: `StateMachine`, `Configuration`, and `*AirGradient`. 

As you say we can pass `Configuration`? Or should remove alls?

airgradienthq (approved):

PR #104, opened 2024-05-02
Feature/show ota process on display

pnt325:


examples/OneOpenAir/OtaHandler.h
```diff
@@ -18,10 +21,23 @@ enum OtaUpdateOutcome {
 
 class OtaHandler {
 public:
-  void updateFirmwareIfOutdated(String deviceId) {
+  OtaHandler(StateMachine &sm, Configuration &config)
```
nick-4711, line 24: Propose to use a callback instead of passing in the state machine. Also the deviceId should be sufficient to pass in, haven't checked really though.

examples/OneOpenAir/OtaHandler.h
```diff
@@ -18,10 +21,23 @@ enum OtaUpdateOutcome {
...
+      : sm(sm), config(config) {}
+  void setAirGradient(AirGradient *ag) { this->ag = ag; };
```
nick-4711, line 26: Propose to avoid to pass in *ag.

examples/OneOpenAir/OneOpenAir.ino
```diff
@@ -434,6 +435,25 @@ static bool sgp41Init(void) {
   return false;
 }
 
+static void otaHandlerCallback(StateMachine::OtaState state, String mesasge) {
```
nick-4711, line 438: This way we still have a dependency to the StateMachine. Propose to avoid this.

examples/OneOpenAir/OtaHandler.h
```diff
@@ -19,7 +15,14 @@ enum OtaUpdateOutcome {
   UDPATE_SKIPPED
 };
 
-typedef void(*OtaHandlerCallback_t)(StateMachine::OtaState state,
+enum OtaState {
+  OTA_STATE_BEGIN,
+  OTA_STATE_FAIL,
```
nick-4711, line 20: Propose to rename to FAILED

examples/OneOpenAir/OtaHandler.h
```diff
@@ -19,7 +15,14 @@ enum OtaUpdateOutcome {
...
+  OTA_STATE_PROCESSING,
+  OTA_STATE_SUCCESS
+};
+
+typedef void(*OtaHandlerCallback_t)(OtaState state,
```
nick-4711, line 25: Can we use a unsigned int named "progress" instead of "message"?

examples/OneOpenAir/OtaHandler.h
```diff
@@ -43,7 +46,7 @@ class OtaHandler {
       while (i != 0) {
         i = i - 1;
         if (this->callback) {
-          this->callback(StateMachine::OtaState::OTA_STATE_SUCCESS, String(i));
+          this->callback(OtaState::OTA_STATE_SUCCESS, String(i));
         }
         delay(1000);
```
nick-4711, line 51: I think we shouldn't delay the process in here, this can be moved out.

nick-4711: This is fine but should be done outside the OTA handler

pnt325: Ok

nick-4711 (approved):

PR #107, opened 2024-05-08
Move structure configure to JSON

pnt325:


src/AgConfigure.cpp
```diff
@@ -53,22 +75,17 @@ String Configuration::getLedBarModeName(LedBarMode mode) {
  *
  */
 void Configuration::saveConfig(void) {
```
nick-4711, line 77: Propose to name the file according to the class name, so AgConfigure -> Configuration

src/AgConfigure.cpp
```diff
@@ -81,90 +98,55 @@ void Configuration::saveConfig(void) {
 
 void Configuration::loadConfig(void) {
   bool readSuccess = false;
+  char *buf = (char *)malloc(EEPROM_CONFIG_SIZE);
+  if (buf == NULL) {
+    logError("Malloc read file buffer failed");
```
nick-4711, line 103: we don't really allocated a "file buffer"?

src/AgConfigure.cpp
```diff
@@ -81,90 +98,55 @@ void Configuration::saveConfig(void) {
...
+    buf[i] = EEPROM.read(i);
   }
   readSuccess = true;
 #else
   File file = SPIFFS.open(CONFIG_FILE_NAME);
   if (file && !file.isDirectory()) {
-    if (file.readBytes((char *)&config, sizeof(config)) == sizeof(config)) {
-      readSuccess = true;
-    }
+    logInfo("Read file");
```
nick-4711, line 115: Since we are  about to read the file I propose to change the log message to
"Reading file"

src/AgConfigure.cpp
```diff
@@ -81,90 +98,55 @@ void Configuration::saveConfig(void) {
...
+    file.readBytes(buf, file.size());
```
nick-4711, line 116: Error handling missing?

src/AgConfigure.cpp
```diff
@@ -81,90 +98,55 @@ void Configuration::saveConfig(void) {
...
+    readSuccess = true;
     file.close();
-  }
-#endif
-
-  if (!readSuccess) {
-    logError("Load configure failed");
```
nick-4711, line 101: Propose to write

Loading configuration failed, using default configuration.

src/AgConfigure.cpp
```diff
@@ -900,11 +864,255 @@ String Configuration::getAbcDayString(int value) {
...
+    saveConfig();
+  }
+}
+
 String Configuration::getFailedMesage(void) { return failedMessage; }
 
 void Configuration::setPostToAirGradient(bool enable) {
-  if (enable != config.postDataToAirGradient) {
-    config.postDataToAirGradient = enable;
+  bool oldEnabled = jconfig[jprop_postDataToAirGradient];
```
nick-4711, line 1113: I don't understand here what "old" means

src/AgConfigure.cpp
```diff
@@ -972,13 +1186,15 @@ int Configuration::getLastOta(void) {
     logError("Current year " + String(curYear) + String(" invalid"));
     return -1;
   }
+  double _t = jconfig[jprop_lastOta];
+  time_t lastOta = (time_t)_t;
```
nick-4711, line 1190: I think we can avoid to work with time values for the OTA feature, and use some measurement cycle counter instead.

src/AgConfigure.h
```diff
@@ -8,31 +8,6 @@
 
 class Configuration : public PrintLog {
```
nick-4711, line 9: There is probably a better solution for the inheritance here, maybe create a new issue to look into this.

src/AgConfigure.cpp
pnt325: Ok

src/AgConfigure.cpp
pnt325: I have check JSON value on `toConfig`. If reading file failed or the JSON file invalid that should set as default value

src/AgConfigure.cpp
pnt325: `old` is current configuration valu. It's use to compare to new value.

src/AgConfigure.cpp
pnt325: It's make sure after special period device can trigger OTA again if failed. if you're use the counter it's reset after device reset, for multiple OTA can make damage to FLASH and device can not work any more.

src/AgConfigure.cpp
```diff
@@ -109,13 +126,15 @@ void Configuration::loadConfig(void) {
...
   }
-  readSuccess = true;
 #else
   File file = SPIFFS.open(CONFIG_FILE_NAME);
   if (file && !file.isDirectory()) {
-    logInfo("Read file");
-    file.readBytes(buf, file.size());
-    readSuccess = true;
+    logInfo("Reading file...");
+    if(file.readBytes(buf, file.size()) != file.size()) {
```
nick-4711, line 133: code format

src/AgConfigure.cpp
```diff
@@ -43,6 +43,24 @@ JSON_PROP_DEF(co2CalibrationRequested);
 JSON_PROP_DEF(ledBarTestRequested);
 JSON_PROP_DEF(lastOta);
 JSON_PROP_DEF(offlineMode);
+
+#define jprop_model_default                 ""
+#define jprop_country_default               ""
+#define jprop_pmStandard_default            getPMStandardString(false)
+#define jprop_ledBarMode_default            getLedBarModeName(LedBarMode::LedBarModeCO2)
```
nick-4711, line 50: I think we should use upper case here?

nick-4711, line 50: Propose to rename getLedBarModeName -> getLedBarModeAsString

src/AgConfigure.cpp
```diff
@@ -43,6 +43,24 @@ JSON_PROP_DEF(co2CalibrationRequested);
...
+#define jprop_displayMode_default           getDisplayModeString(true)
```
nick-4711, line 51: getDisplayModeString ->  getDisplayModeAsString

nick-4711 (approved):

PR #108, opened 2024-05-09
Display show message for offline/online mode

pnt325:
If current `offline` mode, message show `Press now for online mode` button pressed on 3sec will set to `online` mode.

If current `online` mode, message show `Press now for offline mode` button pressed on 3sec will set to `offline` mode.

If `offline` mode the WiFi connector will not trigger and otherwise.

@airgradienthq  Please consider approve PR  107 first. Because it's use the configuration data save on JSON file.


examples/OneOpenAir/OneOpenAir.ino
```diff
@@ -175,8 +174,30 @@ void setup() {
...
+      if (ag->button.getState() == ag->button.BUTTON_PRESSED) {
+        configuration.setOfflineMode(!configuration.isOfflineMode());
+
+        oledDisplay.setText(
+        "Offline Mode",
+        configuration.isOfflineMode() ? " = True" : "  = False", "");
+        delay(1000);
+        break;
+      }
+      uint32_t ms = (uint32_t)(millis() - stime);
```
dominik-cnx, line 194: propose to give the variable a descriptive name

dominik-cnx (approved):

PR #127, opened 2024-05-15
Hotfix/purple color

pnt325:


src/AgStateMachine.cpp
```diff
@@ -7,6 +7,12 @@
 
 #define SENSOR_CO2_CALIB_COUNTDOWN_MAX 5 /** sec */
 
+#define RGB_COLOR_R 255, 0, 0     /** Red */
```
nick-4711, line 10: R -> RED, ....

airgradienthq: Both. Refactoring and fixing CO2 color for purple.

PR #129, opened 2024-05-15
Turn LED indicate wifi failed, cloud failed and get cloud configuration failed to off

pnt325:


examples/OneOpenAir/OneOpenAir.ino
```diff
@@ -874,13 +874,15 @@ static void configUpdateHandle() {
...
-    state = AgStateMachineWiFiLost;
-  } else if (apiClient.isFetchConfigureFailed()) {
-    stateMachine.displaySetAddToDashBoard();
-    state = AgStateMachineSensorConfigFailed;
-  } else if (apiClient.isPostToServerFailed()) {
-    state = AgStateMachineServerLost;
+  if (configuration.isOfflineMode() == false) {
+    if (wifiConnector.isConnected() == false) {
+      state = AgStateMachineWiFiLost;
+    } else if (apiClient.isFetchConfigureFailed()) {
```
nick-4711, line 880: propose to name the method isFetchConfigurationFailed() or only fetchConfigurationFailed() would be fine too

examples/OneOpenAir/OneOpenAir.ino
```diff
@@ -874,13 +874,15 @@ static void configUpdateHandle() {
...
+      stateMachine.displaySetAddToDashBoard();
```
nick-4711, line 881: I haven't looked at the implementation of isFetchConfigurationFailed() but in this context I can think of these cases where the message can be misleading

1) The fetch failed because of some internet issue, e.g. timeout
2) The fetch failed because of server overload or backend down (e.g. 503 is returned)

I think we should display this message only if the server returns a 400

PR #130, opened 2024-05-15
Fix print number value wrong format

pnt325:


examples/OneOpenAir/OneOpenAir.ino
```diff
@@ -972,7 +972,7 @@ static void updatePm(void) {
       Serial.printf("[1] Relative Humidity: %d\r\n", measurements.hum_1);
       Serial.printf("[1] Temperature compensated in C: %0.2f\r\n",
                     ag->pms5003t_1.temperatureCompensated(measurements.temp_1));
-      Serial.printf("[1] Relative Humidity compensated: %d\r\n",
+      Serial.printf("[1] Relative Humidity compensated: %f\r\n",
```
nick-4711, line 975: we could also define the number of fraction digits, I think one would be enough for hum, but fine

nick-4711 (approved):

PR #132, opened 2024-05-15
Update configuration log message

pnt325:


src/AgConfigure.cpp
```diff
@@ -630,7 +630,7 @@ bool Configuration::parse(String data, bool isLocal) {
     saveConfig();
     printConfig();
   } else {
-    logInfo("Nothing changed ignore udpate");
+    logInfo("Update ignored due to local unofficial changes");
```
nick-4711, line 633: I'm wondering where the case "we are on the latest version" is.

airgradienthq (approved):

PR #137, opened 2024-05-17
Correct OTA process message show on display

pnt325:


examples/OneOpenAir/OneOpenAir.ino
```diff
@@ -519,6 +519,26 @@ static void displayExecuteOta(OtaState state, String msg, int processing) {
     delay(2500);
     break;
   }
+  case OtaState::OTA_STATE_SKIP: {
+    if (ag->isOne()) {
+      oledDisplay.showNewFirmwareSkipped();
+    } else {
+      Serial.println("Firmware update: Skipped");
```
nick-4711, line 526: Propose to rephrase to
Firmware update skipped

examples/OneOpenAir/OneOpenAir.ino
```diff
@@ -519,6 +519,26 @@ static void displayExecuteOta(OtaState state, String msg, int processing) {
...
+    }
+
+    delay(2500);
+    break;
+  }
+  case OtaState::OTA_STATE_UP_TO_DATE: {
+    if (ag->isOne()) {
+      oledDisplay.showNewFirmwareUpToDate();
+    } else {
+      Serial.println("Firmware update: up to date");
```
nick-4711, line 536: Propose to rephrase to
Firmware is up to date

examples/OneOpenAir/OneOpenAir.ino
```diff
@@ -529,6 +549,11 @@ static void displayExecuteOta(OtaState state, String msg, int processing) {
     break;
   }
   case OtaState::OTA_STATE_SUCCESS: {
+    if (ag->isOne()) {
+      oledDisplay.showNewFirmwareUpdating(String(100));
```
nick-4711, line 553: Not related to the PR but I would expect that we can write this here in this context

oledDisplay.showFirmwareUpdateProgress(100);

examples/OneOpenAir/OneOpenAir.ino
```diff
@@ -544,6 +569,7 @@ static void displayExecuteOta(OtaState state, String msg, int processing) {
         
         delay(1000);
       }
+      oledDisplay.setBrightness(0);
```
nick-4711, line 572: Curious to know, what happens without this line?

examples/OneOpenAir/OneOpenAir.ino
```diff
@@ -857,11 +883,12 @@ static void configUpdateHandle() {
   fwNewVersion = configuration.newFirmwareVersion();
   if (fwNewVersion.length()) {
     bool doOta = false;
-    if (measurements.otaBootCount == 0) {
+    if (measurements.otaBootCount < 0) {
       doOta = true;
       Serial.println("First OTA");
```
nick-4711, line 888: I think the message is misleading, the device might have gone through several OTA updates when we arrive here.

src/AgOledDisplay.cpp
```diff
@@ -380,7 +380,33 @@ void OledDisplay::showNewFirmwareFailed(void) {
...
+
+void OledDisplay::showNewFirmwareSkipped(void) {
+  if (isDisplayOff) {
+    return;
+  }
+
+  DISP()->firstPage();
+  do {
+    DISP()->setFont(u8g2_font_t0_16_tf);
+    setCentralText(20, "Firmware Update");
```
nick-4711, line 396: Propose to rephrase to
Firmware update skipped

src/AgOledDisplay.h
```diff
@@ -40,6 +40,8 @@ class OledDisplay : public PrintLog {
   void showNewFirmwareUpdating(String percent);
```
nick-4711, line 40: propose to rename to
showFormwareUpdateProgress(int percent)

examples/OneOpenAir/OneOpenAir.ino
pnt325: It's relate to PR. Because firmware process has finish 100% but display sometime do not show 100%.

examples/OneOpenAir/OneOpenAir.ino
pnt325: This like will be clear all display content while reboot device. If not the display still Show "Firmware update success" while rebooting

src/AgOledDisplay.h
pnt325: Ok

src/AgOledDisplay.cpp
pnt325: If it's fit for a line we should reduce the font size.

examples/OneOpenAir/OneOpenAir.ino
pnt325: It's indicate first OTA check. If it's zero and process failed, retry will call for 3-4 times (Because we check cloud configure each 15sec). < 0 should help, that mean the otaBootCount is invalid, after first check it's set to bootCount

examples/OneOpenAir/OneOpenAir.ino
nick-4711: Ehm, I mean that we should pass the value along as an int (and not a String)

nick-4711 (approved):

PR #143, opened 2024-05-18
Reboot After 180s WiFi Manager not Connect.

pnt325:


MallocArray:
What about the case where a device is moved to a new location just to measure, such as when I bring mine to other families homes to get a sense of what their air quality is.  I intentionally don't connect it to their wifi, but I wouldn't want the device to be in a constant boot loop, especially since the VOC/NOx sensor would be basically useless in this case, since the baseline would keep getting reset every time the device rebooted due to no Wifi connection.

airgradienthq:
For this we have the new offline mode. Then the monitor will just display the metrics on the display but not trying to connect to any WiFi network etc.
To enable the offline mode, use fw version 3.1.1. When booting it will show a message "Press for offline mode". Then press the button once. This will be persistent on the device.

MallocArray:
What does it take to get it out of Offline mode?

I have done the use case I mentioned, bringing my home device somewhere else for a weekend.  If I put it into Offline Mode while I was away, would I have to flash it again to get it out of Offline mode, or is there an option to press the button during bootup to come out of Offline mode?

airgradienthq:
Exactly. Just press the button again on boot and it will go back into the normal (WiFi) mode.

nick-4711 (approved):

PR #144, opened 2024-05-18
Feature/show status icon on display

pnt325:
disp_cloud_issue
disp_wifi_signal

### DIFF
--- a/src/AgOledDisplay.cpp
+++ b/src/AgOledDisplay.cpp
@@ -85,6 +85,16 @@ const xbm_icon_t cloud_icons[] = {
   [CLOUD_ICON_ISSUE] = {.width = cloud_issue_width, .height = cloud_issue_height, .icon = cloud_issue_bits},
 };
 
+#define offline_width  14
+#define offline_height 11
+static unsigned char offline_bits[] = {
+ 0xe9,0xc7,0xe6,0xdf,0x06,0xf8,0xe9,0xe3,0xe0,0xcf,0x0c,0xcc,
+ 0xe0,0xc1,0xf0,0xc3,0x00,0xc0,0xc0,0xc0,0xc0,0xc0};
+
+ const xbm_icon_t offline_icon[] = {
+  {.width = offline_width, .height = offline_height, .icon = offline_bits},
+ };
+
 /** Cast U8G2 */
 #define DISP() ((U8G2_SH1106_128X64_NONAME_F_HW_I2C *)(this->u8g2))
 
@@ -345,6 +355,12 @@ void OledDisplay::showDashboard(OledDisplay::DashboardStatus status) {
     }
     case DashBoardStatusDeviceId: {
       setCentralText(10, ag->deviceId().c_str());
+      break;
+    }
+    case DashBoardStatusOfflineMode: {
+      showIcon(icon_pos_x, 0, &offline_icon[0]);
+      showTempHum(true);
+      break;
     }
     default:
       break;
--- a/src/AgOledDisplay.cpp
+++ b/src/AgOledDisplay.cpp
@@ -1,5 +1,85 @@
 #include "AgOledDisplay.h"
 #include "Libraries/U8g2/src/U8g2lib.h"
+#include "WiFi.h"
+
+typedef struct {
+  const int width;
+  const int height;
+  const unsigned char* icon;
+} xbm_icon_t;
+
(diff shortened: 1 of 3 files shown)


src/AgOledDisplay.cpp
```diff
@@ -1,5 +1,89 @@
...
+  const unsigned char* icon;
+} xbm_icon_t;
+
+/** WiFi ICON */
+enum {
+  WIFI_ICON_ISSUE,
+  WIFI_ICON_25,
+  WIFI_ICON_50,
+  WIFI_ICON_75,
+  WIFI_ICON_100,
```
nick-4711, line 21: is 100 and MAX not the same?

src/AgOledDisplay.cpp
```diff
@@ -1,5 +1,89 @@
...
+  [WIFI_ICON_75] = {.width = wifi_75_width, .height = wifi_75_height, .icon = wifi_75_bits},
+  [WIFI_ICON_100] = {.width = wifi_100_width, .height = wifi_100_height, .icon = wifi_100_bits},
+};
+
+/** Cloud ICON */
+enum
+{
+  CLOUD_ICON_NORMAL,
+  CLOUD_ICON_ISSUE,
+  CLOUD_ICON_MAX,
```
nick-4711, line 68: Hmm, what means MAX?

src/AgOledDisplay.cpp
```diff
@@ -219,23 +307,31 @@ void OledDisplay::showDashboard(const char *status) {
...
+      showIcon(icon_pos_x, 0, &wifi_icons[icon_type]);
     } else {
-      String strStatus = "Show status: " + String(status);
-      logInfo(strStatus);
-
-      int strWidth = DISP()->getStrWidth(status);
-      DISP()->drawStr((DISP()->getWidth() - strWidth) / 2, 10, status);
-
-      /** Show WiFi NA*/
       if (strcmp(status, "WiFi N/A") == 0) {
```
nick-4711, line 331: Propose to introduce an enum and avoid implementation of business logic based on strings that are possibly localized in the future / shown to the user / subject to change.

src/AgOledDisplay.cpp
pnt325: In normal develop I define _MAX  on `enum` use to define the buffer size or validate value. Like define array of icon bellow. If I don't use `[index] = {value}` I should use `wifi_icons[WIFI_ICON_MAX]`
```
const xbm_icon_t wifi_icons[] = {
  [WIFI_ICON_ISSUE] = {.width = wifi_issue_width, .height = wifi_issue_height, .icon = wifi_issue_bits},
  [WIFI_ICON_25] = {.width = wifi_25_width, .height = wifi_25_height, .icon = wifi_25_bits},
  [WIFI_ICON_50] = {.width = wifi_50_width, .height = wifi_50_height, .icon = wifi_50_bits},
  [WIFI_ICON_75] = {.width = wifi_75_width, .height = wifi_75_height, .icon = wifi_75_bits},
  [WIFI_ICON_100] = {.width = wifi_100_width, .height = wifi_100_height, .icon = wifi_100_bits},
};
```

MAX mean max item of icons

src/AgOledDisplay.cpp
pnt325: Ok, I will create global string instead. Because we are also meed to string message `Add to dashboard` along side with icons

nick-4711: Still think that we need an enum

src/AgOledDisplay.cpp
nick-4711: Can we remove it?

src/AgOledDisplay.cpp
pnt325: OK

pnt325:
@nick-4711
There are some update. Could you please review it?

nick-4711:
> @nick-4711
> There are some update. Could you please review it?

there are still pending issues?

pnt325:
> > @nick-4711
> > There are some update. Could you please review it?
> 
> there are still pending issues?

You should verify and let me know if any problem if not please approve it.

airgradienthq:
I think we probably also need some icon for the offline mode.

pnt325:
@airgradienthq 
Add offlineMode icon

image

PR #149, opened 2024-05-22
Official OTA 3.1.1 version not reporting FW version on dashboard

pnt325:


examples/OneOpenAir/OneOpenAir.ino
```diff
@@ -861,7 +864,9 @@ static void configUpdateHandle() {
       doOta = true;
       Serial.println("First OTA");
     } else {
-      if ((measurements.bootCount - measurements.otaBootCount) >= 30) {
+      /** Only check for update each 2h*/
+      const float otaBootCount = 120.0f / (SERVER_SYNC_INTERVAL / 60000.0f);
```
nick-4711, line 868: We need to be more frequent than every 2h, propose to make it 1h to be on the safe side.

nick-4711 (approved):

PR #155, opened 2024-05-29
Handle LED bar test button before init sensor initialize

pnt325:


examples/OneOpenAir/OneOpenAir.ino
```diff
@@ -623,6 +623,23 @@ static void oneIndoorInit(void) {
   ag->button.begin();
   ag->watchdog.begin();
 
+  /** Run LED test on start up */
```
nick-4711, line 626: Propose to say

Run LED test on start up if button pressed

nick-4711 (approved):

Issue #161, opened 2024-06-04
LED inconsistency | Firmware v. 3.1.1


yegormk:
I've noticed some issues with the LED bar:
The pictures might not be very clear, but you can see green lights on the first and last LED lights, and in the middle, they are all yellow.
Maybe all LEDs should be yellow and not: "green | yellow | yellow | yellow | green" ? 

![Ticket1](https://github.com/airgradienthq/arduino/assets/82291871/d4efccdb-4096-4bf5-bdf8-3407f37ff09e)
![Ticket2](https://github.com/airgradienthq/arduino/assets/82291871/0464ed72-cc2d-4fb6-8221-02864fb9e9f9)

There also could be some issues on other levels.

pnt325:
@yegormk 
As for your  CO2 value and number of LED is turn off the LED color should be `Organge`

yegormk:
@pnt325 Yes, that is correct. 
But the colors on the first and last LEDs lamps are _Green_ and not _Orange_.
Maybe these two also should be _Orange_?

pnt325:
> @pnt325 Yes, that is correct. But the colors on the first and last LEDs lamps are _Green_ and not _Orange_. Maybe these two also should be _Orange_?

Thank for your report. I'll check it.

airgradienthq:
@yegormk Could you please check if that issue still exists with fw 3.1.4?

yegormk:
@airgradienthq @pnt325 
Issue is still there 
![image](https://github.com/airgradienthq/arduino/assets/82291871/48ebe41a-ccfb-41cc-86d3-3286fbd60909)
![image](https://github.com/airgradienthq/arduino/assets/82291871/6ae4e664-06fe-4e9b-99b9-b285a22a369c)

I have Firmware version 3.1.4:
![image](https://github.com/airgradienthq/arduino/assets/82291871/9cdbead5-a6e8-4ef1-92f7-2e2d86816d55)

Could you please check on your devices?

pnt325:
hi @yegormk 
Thank for your detail report. I have update color code #196 it look better now.

PR #196, opened 2024-07-08
Correct LED bar show sensor value level color

pnt325:
LED bar show `orange`  and `purple` incorrect. #161

### DIFF
--- a/src/AgStateMachine.cpp
+++ b/src/AgStateMachine.cpp
@@ -9,9 +9,9 @@
 
 #define RGB_COLOR_R 255, 0, 0    /** Red */
 #define RGB_COLOR_G 0, 255, 0    /** Green */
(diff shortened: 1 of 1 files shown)


airgradienthq:
@pnt325 What is the reason for this change?

image

Also seems the colors are wrongly defined but were correct before.

Google Says Orange:
image

Now in the code it seems it was correct before but not anymore. I think same applies for the other colors.

pnt325:
> @pnt325 What is the reason for this change?
> 
> image
> Also seems the colors are wrongly defined but were correct before.
> 
> Google Says Orange: image
> 
> Now in the code it seems it was correct before but not anymore. I think same applies for the other colors.
> 
> Also, please correct the comment. Orange is written wrongly.

Yes, Last time I'm also use google recommend value but it's look not correct color. Can you run the code to see what differences?

Issue #190, opened 2024-06-27
Ignore parameter values out of range


airgradienthq:
It seems the PMS5003T seldomly sends extremely low values.

See:
https://github.com/home-assistant/core/issues/120586

I would suggest when reading values from sensor we accept only the following ranges:

Temperature: -40°C to 125°C
Relative Humidity: 0% - 100%
CO2: 0-10.000ppm
PM1, 2.5, 10: 0 - 1000 ug/m3

Byter09:
There seems to be a pattern to this though. Every time my temperature goes to `-1000.99`, the PM sensors also go to `-1 µg/m³` or `particles/dL` for the 0.3 PM sensor. The humidity sensor also goes to -1%.

I do not observe this behaviour with the CO2 value, however I'm sure it can't hurt to limit it to sane values as well :)


What I'm trying to say is that maybe there's something that triggers these values to happen at the same time. Let me know if there's anything I can do to help diagnose this.

airgradienthq:
@Byter09 Yes that makes sense. The temperature/humidity in the outdoor monitor comes from the PM sensor. So that PM/temperature/humidity fails at the same time indicates a com problem.

Byter09:
Here's one more weird quirk I noticed. My internet was off over the night (maintenance by my ISP) until a few minutes ago.

In that whole time, instead of spiking once per day, it spiked several times. I have the feeling this might somehow be related to it not being able to reach the server? Maybe it restarts and the initial value is weird? I don't know. Just leaving this here in the hope that it helps

![Screenshot_2024-07-01-08-32-49-80_c3a231c25ed346e59462e84656a70e50](https://github.com/airgradienthq/arduino/assets/16060697/11a4d6e6-356b-437e-90f6-9b851c001e9e)

mhlas7:
Adding another data point here, I turned off "Post Data to AirGradient" two days in Home Assistant (my sensor is set to local control, not cloud) and I now see the temp read -840 F for a two hour period at least once per day. Before I turned it off, I was getting one off spikes down to -840 F similar to other users in this thread. This makes me think the post data to AirGradient function has some sort of impact on the temp reading.

![image](https://github.com/airgradienthq/arduino/assets/3398654/4c61e8cb-d071-401d-815b-8bcb6a79de77)

gslender:
[...]

PR #197, opened 2024-07-10
Ignore parameter values out of range

pnt325:
Correct sensor value has out of range #190


src/Main/utils.cpp
```diff
@@ -0,0 +1,49 @@
...
+    return 0;
+  }
+  if (value > 10000) {
+    return 10000;
+  }
+  return value;
+}
+
+int utils::correctPMS(int value) {
+  if (value < 10) {
```
Byter09, line 42: Why not allow values below 10? My sensors regularly go below that. Maybe limit it to 0?

pnt325: @Byter09 Please check the issue #190 and add your comments at there.

Byter09: I mean I don't have to. Even that issue says 0 should be the lowest value for PM data.

airgradienthq (changes requested):
A few things:
Only negative PM values (below 0) need to be ignored. 0..10 are valid readings and should be allowed.

Also, if a value is outside the range, the value should be completely ignored.

So for example Temperature -100 should NOT return -40 but be ignored. This would apply to all checks.

pnt325:
> A few things: Only negative PM values (below 0) need to be ignored. 0..10 are valid readings and should be allowed.
> 
> Also, if a value is outside the range, the value should be completely ignored.
> 
> So for example Temperature -100 should NOT return -40 but be ignored. This would apply to all checks.

For this way we don't need to correct value in range, just need to verify it's valid or not before use(show on display or sync to cloud). 

The correction method will be remove and replace check valid valid method. 

```c
bool utils::validTemperature(float value) {
  if (value >= VALID_TEMPERATURE_MIN && value <= VALID_TEMPERATURE_MAX) {
    return true;
  }
  return false;
}
```

@airgradienthq  How do you think?

airgradienthq:
@pnt325 yes I think that's how it should work.

Byter09 (commented):
I know I was asked to review this, but I don't know internals. So from a basic look at it, it looks good to me. :stuck_out_tongue:

airgradienthq (approved):

PR #198, opened 2024-07-12
Add combined `hybrid` led bar mode

stablestud:
This commit adds an additional led mode for the AirGradient One devices. Its called `hybrid` and it combines the CO2 and PM2.5 led modes displaying the worst value of both.

This allows the user to immediately evaluate the current air quality with a single glance. CO2 and PM2.5 can be checked combined directly from it.

### DIFF
--- a/src/AgConfigure.cpp
+++ b/src/AgConfigure.cpp
@@ -20,6 +20,7 @@ const char *LED_BAR_MODE_NAMES[] = {
     [LedBarModeOff] = "off",
     [LedBarModePm] = "pm",
     [LedBarModeCO2] = "co2",
+    [LedBarModeHybrid] = "hybrid",
 };
 
 #define JSON_PROP_NAME(name) jprop_##name
@@ -81,6 +82,8 @@ String Configuration::getLedBarModeName(LedBarMode mode) {
     return String(LED_BAR_MODE_NAMES[LedBarModePm]);
   } else if (mode == LedBarModeCO2) {
     return String(LED_BAR_MODE_NAMES[LedBarModeCO2]);
+  } else if (mode == LedBarModeHybrid) {
+    return String(LED_BAR_MODE_NAMES[LedBarModeHybrid]);
   }
   return String("unknown");
 }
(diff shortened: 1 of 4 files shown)


stablestud:
Hi @pnt325 @airgradienthq I would like your input on this :)

Is such a feature in your view desired? For me it certainly is, because, as described, it gives the user direct information about the air quality across different air parameters

stablestud:
One thing that could be improved are the LED levels (number of LEDs to show):
PM2.5 and CO2 have different count of LED levels and different colors:
|Led level|CO2|PM2.5|
|---------|---|-----|
|1|G|G|
|2|GG|GG|
|3|YYY|YYY|
|4|OOOO|YYYY|
|5|OOOOO|OOOOO|
|6|RRRRRR|OOOOOO|
|7|RRRRRRR|RRRRRRR|
|8|PPPPPPPP|RRRRRRRR|
|9|PRPRPRPRP|PPPPPPPPP|
|10|  |PRPRPRPRP|

stablestud:
Before this feature can be added the integration for HA and AG dashboard must also be updated first, as in my testing the HA integration fails if it receives a unknown `ledBarMode` value (`hybrid`)

Nevertheless I could successfully switch the led mode with:
```
curl -X PUT http://airsensor.internal/config -H "Content-Type: application/json" --data '{"country":"TH","pmStandard":"ugm3","ledBarMode":"hybrid","abcDays":8,"tvocLearningOffset":12,"noxLearningOffset":12,"mqttBrokerUrl":"","temperatureUnit":"c","configurationControl":"local","postDataToAirGradient":false,"ledBarBrightness":100,"displayBrightness":100,"offlineMode":false}'
```

pnt325:
hi @stablestud 
Thank for your info.

@airgradienthq  please check this propose.

airgradienthq:
@stablestud thank you for this contribution. 

I think we need more discussion on this as any there have been other suggestions on this before. Some for example suggesting to show both PM and CO2 at the same time. Like user 5 LEDs for PM and 5 for CO2 etc.

Others wanted to add VOCs but some don't.

So I think there are many ideas out there that need to be discussed first. Because it is not easy to see on the monitor which mode is set, so the more modes we introduce the more confusion could be caused.

I will see if I can come up with a couple of suggestions and then the community can discuss in more detail.

pnt325:
Thanks @stablestud 
I will close the PR, feel free to open it if any update.

PR #201, opened 2024-07-21
Reintroduce 'ROOTAPI' so domain and protocol can be configured

McJoppy:
eg. setter for api root added as comment in examples/BASIC/BASIC.ino


airgradienthq:
@pnt325 Is that change also in the other example files?

pnt325:
> @pnt325 Is that change also in the other example files?

Yes, all example that use http client need to update. This PR update for `BASIC.ino` only

McJoppy:
I only have the ability to test on hardware using the basic example which is why I only updated that.

I'll add the changes to the other examples too.

airgradienthq:
@McJoppy Are you updating the pull request with the other examples too?

McJoppy:
I've now included comments in each example for changing APIROOT to align with `BASIC.ino`

PR #205, opened 2024-07-29
Send PMS5003T firmware version to Ag Cloud

pnt325:
Add JSON `PMS5003T_FwVer` on data send to cloud

- Single channel:
```json
{
  "PMS5003T_FwVer": 2
}
```

- Two channel:
```json
{
  "channels":{
    "1":{
      "PMS5003T_FwVer": 2
    },
    "2":{
      "PMS5003T_FwVer": 2
    }
  }
}
```

### DIFF
--- a/src/AgValue.cpp
+++ b/src/AgValue.cpp
@@ -4,7 +4,7 @@
 #include "Main/utils.h"
 #include "Libraries/Arduino_JSON/src/Arduino_JSON.h"
 
-#define json_prop_pmsFirmware     "pmsFirmware"
+#define json_prop_pmFirmware     "firmware"
 
 /**
  * @brief Get PMS5003 firmware version string
@@ -68,7 +68,7 @@ String Measurements::toString(bool localServer, AgFirmwareMode fwMode, int rssi,
       }
       if (!localServer) {
 
-        root[json_prop_pmsFirmware] =
+        root[json_prop_pmFirmware] =
             this->pms5003FirmwareVersion(ag->pms5003.getFirmwareVersion());
       }
     }
@@ -160,7 +160,7 @@ String Measurements::toString(bool localServer, AgFirmwareMode fwMode, int rssi,
           }
         }
         if (!localServer) {
-          root[json_prop_pmsFirmware] =
+          root[json_prop_pmFirmware] =
               pms5003TFirmwareVersion(ag->pms5003t_1.getFirmwareVersion());
         }
       }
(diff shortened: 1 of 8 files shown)


nick-4711:
a)

maybe we should name the property

firmware

and prefix the value with PMS5003T, e.g.

{
  "firwmare": "PMS5003T-2"
}

the advantage is that we can then also use this with other sensors

b)
what happen is a PMS5003 is plugged in?

pnt325:
> b) what happen is a PMS5003 is plugged in?

It's only report the PMS5003T firmware if device detected is OpenAir.

nick-4711:
OK, this is fine, but we still have to models

PMS5003T
PMS5003

What happens if PMS5003 is connected?

airgradienthq:
@pnt325 As per latest documentation 5003 also has firmeware version:

[PMS5003 series data manual English V1.3(1).pdf.(1).pdf](https://github.com/user-attachments/files/16492045/PMS5003.series.data.manual.English.V1.3.1.pdf.1.pdf)

So please implement for all models (PMS5003 and PMS5003T)

pnt325:
> @pnt325 As per latest documentation 5003 also has firmeware version:
> 
> [PMS5003 series data manual English V1.3(1).pdf.(1).pdf](https://github.com/user-attachments/files/16492045/PMS5003.series.data.manual.English.V1.3.1.pdf.1.pdf)
> 
> So please implement for all models (PMS5003 and PMS5003T)

Thank @airgradienthq
Look the plantower has document to complex.  I will add the firmware version for both follow the document.

pnt325:
> OK, this is fine, but we still have to models
> 
> PMS5003T PMS5003
> 
> What happens if PMS5003 is connected?

@nick-4711 

There are no special on data package to identify the PMS5003 or PMS5003T. I think it's the firmware version, but not sure because they are also programming newest firmware version for PMS5003 too.

pnt325:
@nick-4711

The report PMS firmware updated:

### Single

```json
{
  "pmsFirmware": "PMS5003-1"
}
```

pnt325:
@nick-4711

The report PMS firmware updated:

### Single

#### PMS5003
```json
{
  "firmware": "PMS5003-1"
}
```

#### PMS5003T
```json
{
  "firmware": "PMS5003T-1"
}
```

## Two channel
```json
{
 "channel":
    "1":{
      "firmware": "PMS5003T-1"
    },
    "2":{
      "firmware": "PMS5003T-1"
    }
}
```

nick-4711:
> > OK, this is fine, but we still have to models
> > PMS5003T PMS5003
> > What happens if PMS5003 is connected?
> 
> @nick-4711
> 
> There are no special on data package to identify the PMS5003 or PMS5003T. I think it's the firmware version, but not sure because they are also programming newest firmware version for PMS5003 too.

I would expect that the T model gives you temp/hum and the other doesn't. Hence there should be a way to identify if we talk to a PMS5003T or PMS5003

nick-4711:
For completeness, can you post how the payload is now with the latest changes, thx.

Issue #207, opened 2024-08-01
Update AQI Breakpoints


MallocArray:
In the ESPHome project, a user brought up that EPA updated the AQI breakpoints earlier this year.

https://github.com/MallocArray/airgradient_esphome/issues/74

Consider updating formula in AirGradient firmware and/or Dashboard to reflect this change, if not done already

https://www.epa.gov/system/files/documents/2024-02/pm-naaqs-air-quality-index-fact-sheet.pdf
https://en.wikipedia.org/wiki/Air_quality_index#Computing_the_AQI

airgradienthq:
We need to synchronize this with an update on our dashboard and also probably change the code for the PM colors on the LEDBar.

MallocArray:
Why would the PM colors be adjusted? If you are calculating the compensated values on the device, then what it sent up to the server and the LED lights should be unaffected, as you are now using a more correct value, but still reflecting the same outcome.

lsemprini:
apropos...this pull request was merged into development but not sure where it got deployed https://github.com/airgradienthq/arduino/pull/208

PR #208, opened 2024-08-01
Update AQI breakpoints to 2024 values

MallocArray:
#207 

Update breakpoints to follow EPA guidelines from 2024

Also correct the formula to follow what is published on Wikipedia which changes a few values by 1 or 0.1 to match the existing breakpoints, such as Index low previously being 50 when the link shows it should be 51

https://www.epa.gov/system/files/documents/2024-02/pm-naaqs-air-quality-index-fact-sheet.pdf
https://en.wikipedia.org/wiki/Air_quality_index#Computing_the_AQI


airgradienthq:
@MallocArray Thank you for contributing this. We will merge it in parallel of the adjustments we will need to do on the AirGradient dashboard side.

airgradienthq:
@MallocArray is there a reason why you have the upper limit capped at 325 ug/m3?
The PMS sensor delivers data up to 500. Shouldn't we increase that number?

`  else if (pm02 <= 325.4)`

airgradienthq:
There is the term "beyond index" by the EPA see for example:
https://2009-2017.state.gov/m/pri/gdi/airquality/c70218.htm

Also here it seems the newer version does not have the 500 as upper cap anymore:
image

https://www.epa.gov/system/files/documents/2024-02/pm-naaqs-air-quality-index-fact-sheet.pdf

I would recommend that we do not cap at 500.

MallocArray:
Thank you for the links.  It also brought me to the AirNow calculator which also does not cap at 500.
https://www.airnow.gov/aqi/aqi-calculator-concentration/

I'll revise the PR to no longer have an upper limit

samuelbles07:
We will merge and make the change to accept values higher more than 500

PR #209, opened 2024-08-07
`round2` support negative value

pnt325:
Update round 2 decimal digit value on JSON [support negative value](https://github.com/DaveGamble/cJSON/issues/175#issuecomment-2253721996)

### DIFF
--- a/src/AirGradient.cpp
+++ b/src/AirGradient.cpp
@@ -41,7 +41,14 @@ String AirGradient::getVersion(void) { return GIT_VERSION; }
 BoardType AirGradient::getBoardType(void) { return boardType; }
 
 double AirGradient::round2(double value) {
-  return (int)(value * 100 + 0.5) / 100.0;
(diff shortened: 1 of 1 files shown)


airgradienthq:
@pnt325 Can you please explain the exact purpose of this code change.

pnt325:
> @pnt325 Can you please explain the exact purpose of this code change.

I use round2 to limit the number of character for double value on JSON. Example  value = `12.3456789`  add to json it's look like
```json
{
"value": 123.456789
}
```

after round
```json
{
"value": 123.46
}
```

It's reduce the data size of JSON. From current code it's just round correct for negative value
https://github.com/airgradienthq/arduino/blob/e46e11c03066a4aba3706b90f9184582c1c94178/src/AirGradient.cpp#L45-L46

if the input value is negative the round value is not correct. This update to correct round negative value like I'm mention to link on description.

Issue #182, opened 2024-06-16
Implement Correction Formula for PM2.5 on Display


airgradienthq:
The PM2.5 value measured by the device can be improved with a formula.

Important: Data send to server should not change because we will use that formula also on our server. So below formula should ONLY affect the PM 2.5 value on the display. It should not be send to the AirGradient server. Only the raw data from PMS module send to AirGradient server.

However, the calculated PM2.5 value should be exposed in the Local Server API as "pm02Compensated". I adjusted the documentation already: https://github.com/airgradienthq/arduino/blob/master/docs/local-server.md

This way, integrations like ESPHome, Home Assistant etc can use it.

If display is switched to display US AQI. The corrected PM2.5 value should be used for the US AQI on the display.

This correction formula splits up the algorithms in bands. Also note that negative values could come out of the formula. In this case, it is recommended to set them as 0.

#### AG<sub>raw</sub> <30:
PM<sub>2.5</sub> = [0.524 x AG<sub>raw</sub>] – [0.0862 x RH] + 5.75

#### 30≤ AG<sub>raw</sub> <50:
PM<sub>2.5</sub> = [0.786 x (AG<sub>raw</sub>/20 - 3/2) + 0.524 x (1 -(AG<sub>raw</sub>/20 - 3/2))] x AG<sub>raw</sub> – [0.0862 x RH] + 5.75

#### 50 ≤ AG<sub>raw</sub> <210:
PM<sub>2.5</sub> = [0.786 x AG<sub>raw</sub>] – [0.0862 x RH] + 5.75

#### 210 ≤ AG<sub>raw</sub> <260:
PM<sub>2.5</sub> = [0.69 x (AG<sub>raw</sub>/50 – 21/5) + 0.786 x (1 - (AG<sub>raw</sub>/50 – 21/5))] x AG<sub>raw</sub> – [0.0862 x RH x (1 - (AG<sub>raw</sub>/50 – 21/5))] + [2.966 x (AG<sub>raw</sub>/50 –21/5)] + [5.75 x (1 - (AG<sub>raw</sub>/50 – 21/5))] + [8.84 x (10<sup>-4</sup>) x AG<sub>raw</sub><sup>2</sup>x (AG<sub>raw</sub>/50 – 21/5)]

#### 260 ≤ AG<sub>raw</sub>:
[...]

PR #212, opened 2024-08-15
correct pm2.5 formula to show value on display

pnt325:
Update the PM2.5 calculate formula #182 . The correct value show on device display or use on local only.

### DIFF
--- a/src/AgOledDisplay.cpp
+++ b/src/AgOledDisplay.cpp
@@ -10,7 +10,7 @@
  * @param hasStatus
  */
 void OledDisplay::showTempHum(bool hasStatus) {
-  char buf[10];
+  char buf[16];
   if (value.Temperature > -1001) {
     if (config.isTemperatureUnitInF()) {
       float tempF = (value.Temperature * 9) / 5 + 32;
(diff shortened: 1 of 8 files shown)


src/AgOledDisplay.cpp
```diff
@@ -304,6 +304,10 @@ void OledDisplay::showDashboard(const char *status) {
       DISP()->drawStr(55, 27, "PM2.5");
 
       /** Draw PM2.5 value */
+      int pm25 = value.pm25_1;
+      if (config.hasSensorSHT) {
+        pm25 = ag->pms5003.pm25Compensated(pm25, value.Humidity);
```
nick-4711, line 309: propose to rename the method to simply .compensate(). We can see from the params that this is related to pm25.

src/PMS/PMS.cpp
```diff
@@ -270,6 +270,41 @@ int PMSBase::pm25ToAQI(int pm02) {
     return 500;
 }
 
+/**
+ * @brief Correction PM2.5
+ * 
+ * @param pm25 Raw PM2.5 value
+ * @param humidity Humidity value (%)
+ * @return float
```
nick-4711, line 278: it returns an int? Propose to write

@return the compensated PM2.5 value

src/PMS/PMS.cpp
```diff
@@ -270,6 +270,41 @@ int PMSBase::pm25ToAQI(int pm02) {
...
 
+/**
+ * @brief Correction PM2.5
+ * 
+ * @param pm25 Raw PM2.5 value
+ * @param humidity Humidity value (%)
+ * @return float 
+ */
+int PMSBase::pm25Compensated(int pm25, float humidity) {
+  float value;
```
nick-4711, line 281: We need to link to the documentation where this is from

src/AgOledDisplay.cpp
pnt325: Ok, I'll change the name and update API comment to mention the input is PM2.5

src/PMS/PMS.cpp
pnt325: Yes, it's return the int. I'll update the method comment

src/PMS/PMS.cpp
pnt325: @airgradienthq  please add your document for this formula.

Issue #221, opened 2024-08-25
Check Display I-9PSL uses PM Compensated Values


airgradienthq:
On the display we should show the compensated values.
pm02Compensated

This should also be used to calculate US-AQI values.

@pnt325 Please confirm that this is implemented.

PR #224, opened 2024-08-25
[Fix] PM2.5 compensated show on display

pnt325:
Update missing add PM2.5 compensated before show on display. #221

### DIFF
--- a/src/AgOledDisplay.cpp
+++ b/src/AgOledDisplay.cpp
@@ -311,7 +311,7 @@ void OledDisplay::showDashboard(const char *status) {
       }
       DISP()->setFont(u8g2_font_t0_22b_tf);
       if (config.isPmStandardInUSAQI()) {
-        if (utils::isValidPMS(pm25)) {
+        if (utils::isValidPm(pm25)) {
           sprintf(strBuf, "%d", ag->pms5003.convertPm25ToUsAqi(pm25));
         } else {
           sprintf(strBuf, "%s", "-");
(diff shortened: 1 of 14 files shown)


src/AgOledDisplay.cpp
```diff
@@ -307,20 +307,21 @@ void OledDisplay::showDashboard(const char *status) {
       int pm25 = value.pm25_1;
       if (config.hasSensorSHT) {
         pm25 = ag->pms5003.compensated(pm25, value.Humidity);
+        logInfo("PM2.5:" + String(value.pm25_1) + String("Compensated:") + String(pm25));
       }
       DISP()->setFont(u8g2_font_t0_22b_tf);
       if (config.isPmStandardInUSAQI()) {
-        if (utils::isValidPMS(value.pm25_1)) {
-          sprintf(strBuf, "%d", ag->pms5003.convertPm25ToUsAqi(value.pm25_1));
+        if (utils::isValidPMS(pm25)) {
```
nick-4711, line 314: the method name isValidPMS is misleading, propose to rename to simply isValid()

src/AgOledDisplay.cpp
```diff
@@ -307,20 +307,21 @@ void OledDisplay::showDashboard(const char *status) {
       int pm25 = value.pm25_1;
       if (config.hasSensorSHT) {
         pm25 = ag->pms5003.compensated(pm25, value.Humidity);
```
nick-4711, line 309: the method name should be renamed to compensate()

src/AgOledDisplay.cpp
pnt325: it's the API on `utils` can not use `isValid` cause there are multiple API check sensor valid like `CO2`, `Temperature`, etc.

nick-4711: OK, then I propose to write

isValidPm()

pnt325: @nick-4711 
Updated 753f22923caa97cf0630c5a8da1ec6c6f44cb47e

PR #232, opened 2024-09-11
Feat/ag-client-timeout

samuelbles07:
## Updates

1. Set default http request timeout to 10s on `AgApiClient` class
2. New `AgApiClient` member to set http request timeout from the caller
3. [Fix](https://github.com/airgradienthq/arduino/blob/master/src/AgApiClient.cpp#L116) `AgApiClient::postToServer` uri formatting not using `apiRoot`

### DIFF
--- a/src/AgApiClient.cpp
+++ b/src/AgApiClient.cpp
@@ -107,9 +107,7 @@ bool AgApiClient::postToServer(String data) {
     return false;
   }
 
-  String uri =
-      "http://hw.airgradient.com/sensors/airgradient:" + ag->deviceId() +
-      "/measures";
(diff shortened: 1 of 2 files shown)


nick-4711:
Can you also post a log where we can see that it works?

samuelbles07:
Sure @nick-4711 . On left terminal is serial monitor from airgradient monitor and on the right side is dummy webserver.

1. _Failed on purpose_ → Give delay for 5s on server side and on client timeout is set to 3s.

failed

2. _Success GET → Delay for 10s on client side and 8s delay on server site for GET request

GET

3. _Success POST → Delay for 10s on client side and 8s delay on server site for POST request

POST

nick-4711 (approved):

pnt325 (approved):

PR #233, opened 2024-09-15
Changed PM initialization to also support the Cubic PM2009X

nick-4711:
Cubic has PM sensors that are compatible with the Plantower PMS5003 series, see https://en.gassensor.com.cn/IndoorPMSensor/list.html.

With this PR we are able to support the PM2009X model.

### DIFF
--- a/examples/OneOpenAir/OneOpenAir.ino
+++ b/examples/OneOpenAir/OneOpenAir.ino
@@ -697,7 +697,7 @@ static void oneIndoorInit(void) {
   ledBarEnabledUpdate();
 
   /** Show message init sensor */
-  oledDisplay.setText("Sensor", "initializing...", "");
+  oledDisplay.setText("Monitor", "initializing...", "");
 
   /** Init sensor SGP41 */
   if (sgp41Init() == false) {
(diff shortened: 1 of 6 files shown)


examples/OneOpenAir/OneOpenAir.ino
```diff
@@ -697,7 +697,7 @@ static void oneIndoorInit(void) {
   ledBarEnabledUpdate();
 
   /** Show message init sensor */
-  oledDisplay.setText("Sensor", "initializing...", "");
+  oledDisplay.setText("Monitor", "initializing...", "");
```
nick-4711, line 700: Not related to the PR, changed this because we then initialize multiple sensors, so we can also say that we initialize the entire monitor.

examples/OneOpenAir/OneOpenAir.ino
```diff
@@ -778,43 +778,43 @@ static void openAirInit(void) {
...
 
-  /** Try to find the PMS on other difference port with S8 */
+  /** Attempt to detect PM sensors */
   if (fwMode == FW_MODE_O_1PST) {
     bool pmInitSuccess = false;
     if (serial0Available) {
       if (ag->pms5003t_1.begin(Serial0) == false) {
         configuration.hasSensorPMS1 = false;
-        Serial.println("PMS1 sensor not found");
+        Serial.println("No PM sensor detected on Serial0");
```
nick-4711, line 787: The idea is to start to use a more general term PM that can refer to sensors from different manufacturers.

pnt325:
@nick-4711 
Are there any specification documents? I check from website still unable to download the datasheet or application manual?

PR #235, opened 2024-09-15
Fix/mqtt-log

samuelbles07:
## Changes

Skip initializing mqtt client when mqtt is not configured on the device

## Result

Screen Shot 2024-09-15 at 22 54 31

### DIFF
--- a/examples/BASIC/BASIC.ino
+++ b/examples/BASIC/BASIC.ino
@@ -276,7 +276,7 @@ static void initMqtt(void) {
   if (mqttClient.begin(mqttUri)) {
     Serial.println("Successfully connected to MQTT broker");
   } else {
-    Serial.println("Connect to MQTT broker failed");
+    Serial.println("Connection to MQTT broker failed");
   }
 }
 
--- a/examples/BASIC/BASIC.ino
+++ b/examples/BASIC/BASIC.ino
@@ -266,10 +266,17 @@ static void mdnsInit(void) {
 }
 
 static void initMqtt(void) {
-  if (mqttClient.begin(configuration.getMqttBrokerUri())) {
-    Serial.println("Setup connect to MQTT broker successful");
+  String mqttUri = configuration.getMqttBrokerUri();
+  if (mqttUri.isEmpty()) {
+    Serial.println(
+        "MQTT is not configured, skipping initialization of MQTT client");
+    return;
+  }
+
+  if (mqttClient.begin(mqttUri)) {
+    Serial.println("Successfully connected to MQTT broker");
   } else {
-    Serial.println("setup Connect to MQTT broker failed");
+    Serial.println("Connect to MQTT broker failed");
   }
 }
 
--- a/examples/DiyProIndoorV3_3/DiyProIndoorV3_3.ino
+++ b/examples/DiyProIndoorV3_3/DiyProIndoorV3_3.ino
@@ -274,7 +274,7 @@ static void initMqtt(void) {
   if (mqttClient.begin(mqttUri)) {
     Serial.println("Successfully connected to MQTT broker");
   } else {
-    Serial.println("Connect to MQTT broker failed");
+    Serial.println("Connection to MQTT broker failed");
   }
 }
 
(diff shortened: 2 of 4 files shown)


examples/OneOpenAir/OneOpenAir.ino
```diff
@@ -382,7 +382,13 @@ static void createMqttTask(void) {
 }
 
 static void initMqtt(void) {
-  if (mqttClient.begin(configuration.getMqttBrokerUri())) {
+  String mqttUri = configuration.getMqttBrokerUri();
+  if (mqttUri.isEmpty()) {
+    Serial.println("MQTT is not configured, skip initialize mqtt client");
```
nick-4711, line 387: propose to rephrase to

.., skipping initialization of mqtt client

and decide if we want to use capitalize or not. Maybe 

"MQTT URL  is not configured, , skipping initialization of MQTT client

examples/OneOpenAir/OneOpenAir.ino
```diff
@@ -382,7 +382,13 @@ static void createMqttTask(void) {
...
+    return;
+  }
+
+  if (mqttClient.begin(mqttUri)) {
     Serial.println("Connect to MQTT broker successful");
```
nick-4711, line 392: Here we could write the past to indicate that it actually worked

"Connected to MQTT broker successfully"

examples/OneOpenAir/OneOpenAir.ino
```diff
@@ -382,6 +382,12 @@ static void createMqttTask(void) {
 }
 
 static void initMqtt(void) {
+  String mqttUri = configuration.getMqttBrokerUri();
+  if (mqttUri.isEmpty()) {
+    Serial.println("MQTT is not configured, skip initialize mqtt client");
```
pnt325, line 387: There are log message in `mqttClient.begin`. Just update the log message inside is good enough. Cause MQTT also usage on ESP8266 device too. 

https://github.com/samuelbles07/arduino/blob/bf1a55793768ff8240201e70ac822b6bcd6f022d/src/MqttClient.cpp#L20-L28

samuelbles07: Yes, but it return false. Hence in the implementation side will print `Connect to MQTT broker failed`. Just realized it is used for others too, will update about this.

pnt325: Don't forget to test with ESP8266 too.

samuelbles07:
Screen Shot 2024-09-16 at 15 22 58
Screen Shot 2024-09-16 at 15 29 34

examples/BASIC/BASIC.ino
```diff
@@ -266,10 +266,17 @@ static void mdnsInit(void) {
...
+    Serial.println(
+        "MQTT is not configured, skipping initialization of MQTT client");
+    return;
+  }
+
+  if (mqttClient.begin(mqttUri)) {
+    Serial.println("Successfully connected to MQTT broker");
   } else {
-    Serial.println("setup Connect to MQTT broker failed");
+    Serial.println("Connect to MQTT broker failed");
```
nick-4711, line 279: Propose to rename to

Connection to MQTT broker failed

examples/DiyProIndoorV3_3/DiyProIndoorV3_3.ino
```diff
@@ -264,10 +264,17 @@ static void mdnsInit(void) {
...
+    Serial.println(
+        "MQTT is not configured, skipping initialization of MQTT client");
+    return;
+  }
+
+  if (mqttClient.begin(mqttUri)) {
+    Serial.println("Successfully connected to MQTT broker");
   } else {
-    Serial.println("setup Connect to MQTT broker failed");
+    Serial.println("Connect to MQTT broker failed");
```
nick-4711, line 277: dito

examples/BASIC/BASIC.ino
samuelbles07: Updated

nick-4711 (approved):

PR #237, opened 2024-09-16
Retry GET/POST data with server if failed

pnt325:
1. Remove check WiFi connected before send POST from `ApiClient` to `.ino`
2. Retry `3` times and period `5` seconds if POST/GET return failed


nick-4711 (changes requested):
I'm against any retries because it could create excessive load on the server when it is already struggling.

PR #252, opened 2024-10-06
Fix default for ledBarMode

kellycampbell:
Fixes what looks like a copy/paste error.


samuelbles07 (approved):

samuelbles07:
Hi @kellycampbell , thank you for the fix! Can you change the PR to develop branch instead of master?

PR #257, opened 2024-10-22
Send all particle counts and standard particle values provided by the PMS sensor

samuelbles07:
### New measurements

- Standard particle of each PM concentration
- Particle count `1.0`, `2.5` and `10`. Except `10` only for indoors monitor.

### New field on json payload

- `pm01Standard` 
- `pm02Standard`
- `pm10Standard`
- `pm01Count`
- `pm25Count`
- `pm10Count`

See `docs/local-server.md` for more detail of each field above

### DIFF
--- a/docs/local-server.md
+++ b/docs/local-server.md
@@ -50,12 +50,18 @@ You get the following response:
 |-----------------------------------|---------|----------------------------------------------------------------------------------------|
 | `serialno`                        | String  | Serial Number of the monitor                                                           |
 | `wifi`                            | Number  | WiFi signal strength                                                                   |
-| `pm01`                            | Number  | PM1 in ug/m3                                                                           |
-| `pm02`                            | Number  | PM2.5 in ug/m3                                                                         |
-| `pm10`                            | Number  | PM10 in ug/m3                                                                          |
+| `pm01`                            | Number  | PM1.0 in ug/m3 (atmospheric environment)                                               |
+| `pm02`                            | Number  | PM2.5 in ug/m3 (atmospheric environment)                                               |
+| `pm10`                            | Number  | PM10 in ug/m3 (atmospheric environment)                                                |
 | `pm02Compensated`                 | Number  | PM2.5 in ug/m3 with correction applied (from fw version 3.1.4 onwards)                 |
+| `pm01_sp`                         | Number  | PM1.0 in ug/m3 (standard particle)                                                     |
+| `pm02_sp`                         | Number  | PM2.5 in ug/m3 (standard particle)                                                     |
+| `pm10_sp`                         | Number  | PM10 in ug/m3 (standard particle)                                                      |
 | `rco2`                            | Number  | CO2 in ppm                                                                             |
-| `pm003Count`                      | Number  | Particle count per dL                                                                  |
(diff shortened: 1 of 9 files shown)


nick-4711:
Propose to give the PR a more specific title. I mean indicate what "other" data is included

docs/local-server.md
```diff
@@ -50,12 +50,18 @@ You get the following response:
...
-| `pm01`                            | Number  | PM1 in ug/m3                                                                           |
-| `pm02`                            | Number  | PM2.5 in ug/m3                                                                         |
-| `pm10`                            | Number  | PM10 in ug/m3                                                                          |
+| `pm01`                            | Number  | PM1.0 in ug/m3 (atmospheric environment)                                               |
+| `pm02`                            | Number  | PM2.5 in ug/m3 (atmospheric environment)                                               |
+| `pm10`                            | Number  | PM10 in ug/m3 (atmospheric environment)                                                |
 | `pm02Compensated`                 | Number  | PM2.5 in ug/m3 with correction applied (from fw version 3.1.4 onwards)                 |
+| `pm01_sp`                         | Number  | PM1.0 in ug/m3 (standard particle)                                                     |
```
nick-4711, line 57: Propose to use camel case and name this pm01Standard

src/PMS/PMS5003T.cpp
```diff
@@ -110,33 +110,68 @@ bool PMS5003T::begin(void) {
...
  *
- * @return int PM 0.3 Count index
+ * @return int particle 0.3 count index
  */
 int PMS5003T::getPm03ParticleCount(void) {
   return pms.getCount0_3();
 }
 
+/**
+ * @brief Read particle 1.0 count must call this function after @ref readData success
```
nick-4711, line 162: Do we really need to call it? Is it not more that we can if we want?

src/AgValue.cpp
```diff
@@ -571,8 +659,9 @@ JSONVar Measurements::buildPMS(AirGradient &ag, int ch, bool allCh, bool withTem
     return pms;
   };
 
-  // Handle both channel with average, if one of the channel not valid, use another one
-  /// PM01
+  /** Handle both channel with average, if one of the channel not valid, use another one */
```
nick-4711, line 662: typos

src/PMS/PMS5003T.cpp
samuelbles07: Yes, totally agree. updated

nick-4711 (approved):

src/AgValue.cpp
```diff
@@ -91,10 +115,34 @@ bool Measurements::update(MeasurementType type, int val, int ch) {
     temporary = &_pm_10[ch];
     invalidValue = utils::getInvalidPmValue();
     break;
+  case PM01_SP:
+    temporary = &_pm_01_sp[ch];
+    invalidValue = utils::getInvalidPmValue();
```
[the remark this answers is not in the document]
samuelbles07: No. `invalidValue` variable act as reference invalid value respective to target measurements. It later check if value provided to `update()` function is an invalid value or not.
Comment added for that variable.

PR #258, opened 2024-10-31
Include PMS particle count 0.5 and 5.0 

samuelbles07:
### New measurements

- Particle count `0.5` and `5.0`. Except `5.0` only for indoor monitor.

### New field on json payload

- `pm005Count`
- `pm50Count`

### Updated field on json payload

- `pm02Count`

See `docs/local-server.md` for more detail of each field above


src/AgApiClient.cpp
```diff
@@ -130,7 +130,7 @@ bool AgApiClient::postToServer(String data) {
   client.end();
 
   logInfo(String("POST: ") + uri);
-  logInfo(String("DATA: ") + data);
+  // logInfo(String("DATA: ") + data);
```
nick-4711, line 133: propose to continue to log the data

src/AgValue.cpp
```diff
@@ -763,6 +795,20 @@ JSONVar Measurements::buildPMS(AirGradient &ag, int ch, bool allCh, bool withTem
     pms["channels"]["2"]["pm003Count"] = ag.round2(_pm_03_pc[1].update.avg);
   }
 
+  /// PM0.5 particle count
+  if (utils::isValidPm03Count(_pm_05_pc[0].update.avg) &&
+      utils::isValidPm03Count(_pm_05_pc[1].update.avg)) {
+    float avg = (_pm_05_pc[0].update.avg + _pm_05_pc[1].update.avg) / 2.0f;
+    pms["pm005Count"] = ag.round2(avg);
+    pms["channels"]["1"]["pm005Count"] = ag.round2(_pm_05_pc[0].update.avg);
```
nick-4711, line 803: 005 vs 05, I think we should align this to avoid confusion

src/PMS/PMS5003.cpp
```diff
@@ -136,13 +136,27 @@ int PMS5003::getPm03ParticleCount(void) {
  */
 int PMS5003::getPm01ParticleCount(void) { return pms.getCount1_0(); }
 
+/**
```
nick-4711, line 139: I think such descriptions don't provide any value, the same information is already included in the method signature. Propose to gradually remove these.

src/AgApiClient.cpp
samuelbles07: We already log `AgValue::toString()` return value [here](https://github.com/airgradienthq/arduino/blob/feat/particle-count/src/AgValue.cpp#L508). So it will not printout when HTTP POST occur, but also for MQTT and local server call.
What do you think?

src/AgValue.cpp
samuelbles07: I just follow the existing formatting. Is it better if we use `pms05Count` and `pms50Count` or `pms5Count`? If yes, shouldn't we need to update other field too then?

samuelbles07: Changes
- `pm25Count` to `pm02Count`
- `pm05Count` to `pm50Count`

Detail on `local-server.md`

PR #267, opened 2024-11-27
(Basic DIY): Fix temperature parsing in OLED and debug for Basic DIY version

jakpor:
1. In Basic DYI version of AirGradient, temperature is displayed incorrectly in the OLED screen.
Current display: `T:241 C`, Correct display `T:24.8 C`.

2. In debug logs, some `Serial.printf` calls pass through `String` type instead of `char *`. This is visible only on longer messages, such as `Temperature`.
Current log: 
`�|�?{1}: 24.17
Humidity{1}: 55.97`
After changes: 
`Temperature{1}: 24.17
Humidity{1}: 55.97`


samuelbles07:
Thanks for the contribution @jakpor !

Issue #160, opened 2024-06-02
[docs] instructions to compile


hestiahacker:
# Problem
Attempting to compile in the Arduino IDE results in the following error:

```
/home/user/air_gradient_open_air_firmware/examples/OneOpenAir/OneOpenAir.ino:40:10: fatal error: AirGradient.h: No such file or directory
   40 | #include "AirGradient.h"
      |          ^~~~~~~~~~~~~~~
compilation terminated.
exit status 1

Compilation error: AirGradient.h: No such file or directory
```

The tactical problem issue here is that some included files are in examples/OneOpenAir, while others are in src, and Arduino IDE is not looking in all these places.

However, the core problem is I can not find instructions on how to configure the Arduino IDE to compile. Since I expect others can compile the code, I expect this is just a setup problem and not a broken codebase.

# Steps to reproduce
1. Set up Arduino IDE to compile for the ESP32 board using the [instructions on the blog](https://www.airgradient.com/blog/install-arduino-c3-mini/)
2. Install the "serial" library for Python (e.g. `sudo apt install python3-serial`)
3. Checkout this repo and open `./examples/OneOpenAir/OneOpenAir.ino` in the Arduino IDE
4. Click the verify button (check mark button in the upper left of the Arduino IDE) and see the posted error above

# Attempts at resolving this
I tried a number of things to resolve this.

- Make a symlink pointing to `./examples/OneOpenAir/OneOpenAir.ino`, put it in `./src/`, and open and compile that (so the file is next to the .h file we want to include)
- Copy `./examples/OneOpenAir/OneOpenAir.ino` into `./src/`, and open and compile that (so the file is next to the .h file we want to include)
- Make a symlink to `./src/Libraries` named `./examples/OneOpenAir/libraries` so the sketch directory has a libraries folder, as described in the [arduino documentation](https://docs.arduino.cc/software/ide-v1/tutorials/installing-libraries/)
- File -> Preference -> Change Sketchbook location to the root git repo, `./src`, or `./examples/OneOpenAir` to get it to find the libraries directory
- copy OneOpenAir.ino into the `./src` directory, rename `./src` to `OneOpenAir`, open OneOpenAir.ino in the Arduino IDE and attempt to compile it (this was tried because Arduino insists on having the name of the directory match the name of the .ino file)

All of the above gave same error message except the last one, which give the error message below:

```
[...]

PR #279, opened 2025-01-25
Update workflow to use compile-sketches action

MallocArray:
Instead of straight CLI, uses the arduino/compile-sketches action.

One immediate benefit is enabling deltas reports
https://github.com/arduino/compile-sketches/tree/v1.1.2/?tab=readme-ov-file#enable-deltas-report

This can show the change in flash and ram compared to the previous run https://github.com/MallocArray/arduino/actions/runs/12954136400/job/36135424360
```
Compiling sketch: examples/OneOpenAir
Compilation time elapsed: 6s
Change in flash: 147290 (7.49%)
Change in RAM for global variables: 2280 (0.7%)
```

A future improvement could be to also add https://github.com/arduino/report-size-deltas that can generate a comment on a PR that also shows the delta information.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -22,7 +22,7 @@ jobs:
             core_url: "https://arduino.esp8266.com/stable/package_esp8266com_index.json"
           - fqbn: "esp32:esp32:esp32c3"
             core: "esp32:esp32"
-            core_version: "2.0.11"
+            core_version: "2.0.17"
             core_url: "https://espressif.github.io/arduino-esp32/package_esp32_index.json"
             board_options: "JTAGAdapter=default,CDCOnBoot=cdc,PartitionScheme=min_spiffs,CPUFreq=160,FlashMode=qio,FlashFreq=80,FlashSize=4M,UploadSpeed=921600,DebugLevel=verbose,EraseFlash=none"
         exclude:
@@ -34,7 +34,7 @@ jobs:
             fqbn: "esp32:esp32:esp32c3"
           - example: "OneOpenAir"
             fqbn: "esp8266:esp8266:d1_mini"
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04
     steps:
       - uses: actions/checkout@v4.2.2
       - uses: arduino/compile-sketches@v1.1.2
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -17,11 +17,14 @@ jobs:
           - "esp32:esp32:esp32c3"
         include:
           - fqbn: "esp8266:esp8266:d1_mini"
-            core: "esp8266:esp8266@3.1.2"
+            core: "esp8266:esp8266"
+            core_version: "3.1.2"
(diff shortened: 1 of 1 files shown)


MallocArray:
Also removed the additional library installation, as current instructions at the top of the example sketches don't mention needing to install any additional libraries

MallocArray:
Feel free to choose another branch if desired. Develop wasn't the default branch, but looks like new code goes to it first, so I submitted it that way

nick-4711 (approved):

MallocArray:
Since this is Approved, can it be completed so the merge finishes?

.github/workflows/check.yml
```diff
@@ -17,11 +17,14 @@ jobs:
           - "esp32:esp32:esp32c3"
         include:
           - fqbn: "esp8266:esp8266:d1_mini"
-            core: "esp8266:esp8266@3.1.2"
+            core: "esp8266:esp8266"
+            core_version: "3.1.2"
             core_url: "https://arduino.esp8266.com/stable/package_esp8266com_index.json"
           - fqbn: "esp32:esp32:esp32c3"
+            core: "esp32:esp32"
+            core_version: "2.0.11"
```
samuelbles07, line 25: We need to set arduino-esp32 version to 2.0.17, since it's the latest version that platformio use. And it's the latest that platformio will be used probably https://github.com/platformio/platform-espressif32/issues/1225

.github/workflows/check.yml
```diff
@@ -33,28 +36,25 @@ jobs:
             fqbn: "esp8266:esp8266:d1_mini"
     runs-on: ubuntu-latest
```
samuelbles07, line 37: Maybe its better to set the ubuntu version

samuelbles07 (approved):

.github/workflows/check.yml
MallocArray: ok, I wrote a note recently into the support e-mail about different locations with different instructions for compiling and currently 
https://github.com/airgradienthq/arduino/blob/54808ac07652c7852fb19da14598f90197b1cdcf/examples/OneOpenAir/OneOpenAir.ino#L18
Only says it was tested against 2.0.11 so I matched it to that.

If you are currently using 2.0.17 when you are developing, this comment should be updated to reflect that as well.

I'll update this line.

.github/workflows/check.yml
MallocArray: I always default and prefer to run the latest until I find that it starts failing because of the version, but I can change it to be ubuntu-24.04 which is the same as latest currently

https://github.com/actions/runner-images

MallocArray:
Addressed the comments

samuelbles07:
Thank you for this @MallocArray !

.github/workflows/check.yml
samuelbles07: Yes, we will update it once we're done with documenting compilation on different environement.

https://github.com/airgradienthq/arduino/issues/160#issuecomment-2661114973

PR #281, opened 2025-02-01
Fix display 0 measurements value on boot

samuelbles07:
### Problem

On version 3.1.13 and 3.1.21, when monitor finish setup() function, display show 0 values for all measurements for a moment.

### Solution

Make sure scheduler for sensor readings run first before display to oled and post to server. By reset display and post data scheduler at the end of setup function.

https://github.com/user-attachments/assets/867325e9-e865-4960-a468-d6d850ab05eb

### DIFF
--- a/examples/OneOpenAir/OneOpenAir.ino
+++ b/examples/OneOpenAir/OneOpenAir.ino
@@ -268,15 +268,17 @@ void setup() {
     oledDisplay.setBrightness(configuration.getDisplayBrightness());
   }
 
-  // Update display and led bar after finishing setup to show dashboard
-  updateDisplayAndLedBar();
+  // Reset display and post schedulers to make sure measurements value already available
+  dispLedSchedule.update();
+  agApiPostSchedule.update();
 }
 
 void loop() {
-  /** Handle schedule */
+  /** Run schedulers */
   dispLedSchedule.run();
   configSchedule.run();
   agApiPostSchedule.run();
+  watchdogFeedSchedule.run();
 
   if (configuration.hasSensorS8) {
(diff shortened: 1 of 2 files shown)


nick-4711:
Propose to add a description, and maybe also a short video that shows how it behaves now.

samuelbles07:
> Propose to add a description, and maybe also a short video that shows how it behaves now.

Description update along with the video

nick-4711:
I can't see the dashes on the video. The idea was to show a dash ('-') whenever a value is unknown. This can be
- right after boot
- when a sensor fails (e.g. if we pull out the S8 then we get a dash for the CO2 value etc.)

samuelbles07:
> I can't see the dashes on the video. The idea was to show a dash ('-') whenever a value is unknown. This can be
> 
> * right after boot
> * when a sensor fails (e.g. if we pull out the S8 then we get a dash for the CO2 value etc.)

Showing dash symbol when sensor values is invalid already integrated on 3.1.9 (AFAIK). But v3.1.13 problem is, when monitor first show dashboard display on the oled, it shows 0 values instead of dash symbol.

On description video, i am making sure measure reading already happen before showing dashboard display so it will directly show real values. But i think set default to invalid value for each measurement on boot would be better. See this [commit](https://github.com/airgradienthq/arduino/pull/281/commits/d2ee3a5d244b13ce08cdebc4044bed59a90765bd).

Results video: https://drive.google.com/file/d/1fVXTNcveLnoolEf77fsaULmuWWnKrBjt/view?usp=drive_link

@nick-4711

nick-4711 (approved):

PR #287, opened 2025-02-24
Integrate how to compile documentation to the codebase

samuelbles07:
Relate to this issue: https://github.com/airgradienthq/arduino/issues/160


docs/howto-compile.md
```diff
@@ -0,0 +1,57 @@
+# How to compile AirGradient firmware on Arduino IDE
+
+## Prequisite
+
+Arduino IDE version 2.x
```
nick-4711, line 5: Propose to name a version that is known to work

docs/howto-compile.md
```diff
@@ -0,0 +1,57 @@
...
+
+## Steps for ESP32C3 based board (ONE and OpenAir Model)
```
nick-4711, line 7: OpenAir -> Open Air

docs/howto-compile.md
```diff
@@ -0,0 +1,57 @@
...
+
+1. Install esp32 board on board manager with version **2.0.17**
+
+![board manager](images/esp32-board.png)
+
+2. Install AirGradient library on library manager using the latest version (at the time of writing, its 3.2.0)
+
+![Aigradient Library](images/ag-lib.png)
+
+3. Plug AirGradient monitor
```
nick-4711, line 17: This should not be needed to only compile?

docs/howto-compile.md
```diff
@@ -0,0 +1,57 @@
...
+JTAG Adapter ➝ Disabled
+Partition Scheme ➝ Minimal SPIFFS (1.9MB APP with OTA/190KB SPIFFS)
+Upload Speed ➝ 921600
+```
+
+![Compile Settings](images/settings.png)
+
+5. Compile
+
+## Steps for ESP8266 based board (DIY model)
```
nick-4711, line 38: Propose to talk about the ONE and Open Air first

docs/howto-compile.md
samuelbles07: Minor version doesn't matter. Already tried with v1, and still work.

docs/howto-compile.md
samuelbles07: Oh yes, you're right!

docs/howto-compile.md
samuelbles07: It already is https://github.com/airgradienthq/arduino/blob/033358e2c25d73e9f7ae1f6a9cf1f2926546bedc/docs/howto-compile.md#steps-for-esp32c3-based-board-one-and-openair-model

The first steps section are for ONE and Open Air

docs/howto-compile.md
```diff
@@ -0,0 +1,55 @@
+# How to compile AirGradient firmware on Arduino IDE
+
+## Prequisite
+
+Arduino IDE version 2.x
```
MallocArray, line 5: Suggest adding a link to download the IDE:
https://www.arduino.cc/en/software

docs/howto-compile.md
```diff
@@ -0,0 +1,55 @@
...
+
+## Steps for ESP32C3 based board (ONE and Open Air Model)
+
+1. Install esp32 board on board manager with version **2.0.17**
```
MallocArray, line 9: I would like this very specific as to which board to include, as there are 2 with ESP32 in the name.  Also how to get to the board manager
Maybe
`1. Install "esp32 by Espressif Systems" in board manager with version **2.0.17** (Tools>Board>Boards Manager)`
image

docs/howto-compile.md
```diff
@@ -0,0 +1,55 @@
...
+
+![board manager](images/esp32-board.png)
+
+2. Install AirGradient library on library manager using the latest version (at the time of writing, its 3.2.0)
```
MallocArray, line 13: Suggestion to leave instructions on how to get to the library manager
`Sketch>Include Library>Manage Libraries...`

docs/howto-compile.md
```diff
@@ -0,0 +1,55 @@
...
+
+```
+Board ➝ ESP32C3 Dev Module
+USB CDC On Boot ➝ Enabled
+CPU Frequency ➝ 160MHz (WiFi)
+Core Debug Level ➝ None (or choose as needed)
+Erase All Flash Before Sketch Upload ➝ Enabled (or choose as needed)
+Flash Frequency ➝ 80MHz
+Flash Mode ➝ QIO
+Flash Size ➝ 4MB
```
MallocArray, line 27: Might as well make this exactly as shown in Arduino IDE to avoid confusion

`Flash Size ➝ 4MB (32Mb)`

docs/howto-compile.md
```diff
@@ -0,0 +1,55 @@
...
+Upload Speed ➝ 921600
+```
+
+![Compile Settings](images/settings.png)
+
+4. Compile
+
+## Steps for ESP8266 based board (DIY model)
+
+1. Add esp8266 board by adding http://arduino.esp8266.com/stable/package_esp8266com_index.json into Additional Board Manager URLs field.
```
MallocArray, line 39: This setting can be difficult to find, since it isn't in the regular Board Manager pane, so instructions on how to get to it would be helpful.  A screenshot would also be helpful.
File>Preferences>Additional boards manager URLs: http://arduino.esp8266.com/stable/package_esp8266com_index.json

docs/howto-compile.md
```diff
@@ -0,0 +1,55 @@
...
+Flash Mode ➝ QIO
+Flash Size ➝ 4MB
+JTAG Adapter ➝ Disabled
+Partition Scheme ➝ Minimal SPIFFS (1.9MB APP with OTA/190KB SPIFFS)
+Upload Speed ➝ 921600
+```
+
+![Compile Settings](images/settings.png)
+
+4. Compile
```
MallocArray, line 35: A large missing piece of this howto is how to get all of the required files from Github.

To try to see things as a new user, I followed these instructions in a new Windows install.  When I got to this step about compiling, it didn't say which file to compile, so I grabbed OneOpenAir.ino from the repo, saved it to my device in Documents/Arduino folder, and it fails with an error about LocalServer.h not available.
image

I downloaded the entire repo as a .zip file and then extracted `arduino-master` to Documents/Arduino folder.  Then opened the file in the IDE `C:\Users\WDAGUtilityAccount\Documents\Arduino\arduino-master\examples\OneOpenAir\OneOpenAir.ino`

This let me compile the firmware exactly as it is, but still relies on the released of the AirGradient library to the Arduino Library manager.  If the reason a user is wanting to compile their own is so they can modify the code, their local IDE needs to be using the local files entirely and not using the published Library.  If I uninstall "AirGradient Air Quality Sensor" from the library manager, I can no longer compile the firmware, as there are files that are not being used by the IDE
image

For me, the only way to do this is to add the repository files to be a subfolder of Arduino/Libraries, so the IDE is using my local files as part of the library so it can find them appropriately.  So instead of putting the contents of the Github repo into Documents/Arduino, I put them in Documents/Arduino/libraries and opened the .ino file and then I was able to compile successfully with all local code, letting me make modifications and have them be used when compiling.
`C:\Users\WDAGUtilityAccount\Documents\Arduino\libraries\arduino-master\examples\OneOpenAir\OneOpenAir.ino`


So I think all of this needs to be covered somewhere and somehow

airgradienthq: @MallocArray thanks for the feedback. Achim here.

When I played around with the Arduino IDE, I saw quite a few restrictions in file locations etc. to get things compiled. I wonder if we can make the assumption that if people use the Arduino IDE they will use the AirGradient Arduino library and install it with the library manager?

If people use other IDE's then everything should work with cloning the repo etc.

Not sure if that makes sense. Just want to throw this into the discussion.

airgradienthq: I agree. I just wanted to point out that the Arduino editor has some strange design choices that are a bit unusual but I'm also not deep enough in that topic anymore. I talked to the team and we will try to address these issues as good as we can.

samuelbles07: I just add another step on how to choose the sketch based on models, this should solve the confusion for not manually download file from the repository.

docs/images/linux-failed.png
MallocArray, line 1: I don't think this file is referenced by the howto-compile.md document any longer and can be removed

docs/howto-compile.md
MallocArray: > I just add another step on how to choose the sketch based on models, this should solve the confusion for not manually download file from the repository.
This is an improvement as it does get it to compile.  
One thing that is not clear is where the files are stored after you install the AirGradient library, as it kind of "just works" but at least in the IDE itself I don't see where it indicates where the sketch is located.

Since an expected use case is users wanting to make their own modifications which could mean needing to modify files in the src folder, at least a mention that the Library files are located by default in the `Documents/Arduino/libraries` for Windows and Mac, and `~/Arduino/libraries` for Linux would be helpful.

docs/howto-compile.md
```diff
@@ -0,0 +1,71 @@
...
+
+## Possible Issues
+
+### Linux (Debian)
+
+ModuleNotFoundError: No module named ‘serial’
+
+Make sure python pyserial module installed globally in the environment by executing: 
+
+`$ pip install pyserial`
```
hestiahacker, line 71: I appreciate this section, however people are going to run into problems installing the library this way.  This should be `sudo apt install -y python3-pyserial` instead.

Here's what happens if I install `sudo apt install -y python3-pip` (it's not installed by default) and then run the command above:

```
$ pip install pyserial
error: externally-managed-environment

× This environment is externally managed
╰─> To install Python packages system-wide, try apt install
    python3-xyz, where xyz is the package you are trying to
    install.
    
    If you wish to install a non-Debian-packaged Python package,
    create a virtual environment using python3 -m venv path/to/venv.
    Then use path/to/venv/bin/python and path/to/venv/bin/pip. Make
    sure you have python3-full installed.
    
    If you wish to install a non-Debian packaged Python application,
    it may be easiest to use pipx install xyz, which will manage a
    virtual environment for you. Make sure you have pipx installed.
    
    See /usr/share/doc/python3.11/README.venv for more information.

note: If you believe this is a mistake, please contact your Python installation or OS distribution provider. You can override this, at the risk of breaking your Python installation or OS, by passing --break-system-packages.
hint: See PEP 668 for the detailed specification.
```

pip does have a `--break-system-packages` option, but as you can imagine, this is generally not a good idea. Making a virtual environment as the error message suggests is one option, but a much simpler option is to use apt to install the package.

MallocArray:
Making sure the comments @hestiahacker made in the Issue are mentioned in this PR

https://github.com/airgradienthq/arduino/issues/160#issuecomment-2689225576

docs/howto-compile.md
samuelbles07: I think it depends on how we install python in our machine. In my case, i'm using [pyenv](https://github.com/pyenv/pyenv) to manage versioning of python so using `apt` will not work (already tried this). On this section, i add to use `apt` as first options with note after.
Thank you for pointing this!

docs/howto-compile.md
samuelbles07: Make sense for pointing this, thanks. i just add a new section on how to contribute to this project. It's still a simple how to, but it should suffice for now.

PR #278, opened 2025-01-24
Fully disable cloud connection to airgradient server option

samuelbles07:
## Changes

1. `disableCloudConnection` config added to disable every cloud connection to airgradient server
2. `disableCloudConnection` not applied for DIY model
3. `local-server.md` updated, describing `offlineMode` and `disableCloudConnection`
4. Adding more verbose logs regarding ignored connection to airgradient server

```
> Ignore fetch server configuration. Either mode is offline or cloud connection disabled or configurationControl set to local
> Ignore send data to server. Either mode is offline or cloud connection is disabled or post data to server disabled
> firmwareCheckForUpdate: mode is offline or cloud connection disabled, ignored
```

PR #290, opened 2025-02-28
fix: don't contact AG server if cloud connection is disabled

hestiahacker:
This should fix the issue where the firmware calls out to the AirGradient server even when "Prevent Connection to AirGradient Server" is checked.

This was brought up in the [ability to turn off data to external networks](https://forum.airgradient.com/t/ability-to-turn-off-data-to-external-networks/3529?u=hestia) post.

It was mentioned in [the forum](https://forum.airgradient.com/t/we-dont-love-subscriptions-but-the-airgradient-dashboard-needs-them/3250/15?u=hestia) that there is an intent to fix this issue.

This change needs tested to verify it completely disables all connections to the AG server. It compiles fine, but I haven't done a full network analysis to confirm it doesn't call out elsewhere or cause any other problem by no longer fetching the server configuration.

### DIFF
--- a/examples/OneOpenAir/OneOpenAir.ino
+++ b/examples/OneOpenAir/OneOpenAir.ino
@@ -872,19 +872,23 @@ void initializeNetwork() {
   checkForUpdateSchedule.update();
 #endif
 
-  apiClient.fetchServerConfiguration();
-  configSchedule.update();
-  if (apiClient.isFetchConfigurationFailed()) {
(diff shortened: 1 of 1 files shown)


samuelbles07:
Hi @hestiahacker , this problem should already fixed on this #278 . You could try the latest firmware version which is 3.2.0.

hestiahacker:
Oops. Sorry about that, closing this PR now

PR #292, opened 2025-03-17
Add cellular connection as network options for AirGradient ONE and Open Air

samuelbles07:
## Changes

1. Support HTTP request using cellular connectivity to communicate to AirGradient server
2. Support firmware over the air update through cellular connectivity 
3. Networking related codes run in dedicated task, that initiated at the end of setup function. Networking codes such as maintain wifi or cellular connectivity, http request to AirGradient server and firmware over the air update
4. Measurement cycle different interval with transmission for cellular network option. With measurement queue to hold every measurement that needs to be sent.
5. Connectivity problem indication from text to icon on oled display for both wifi and cellular
6. CI checkout enable recursive submodule
7. Add [airgradient-client](https://github.com/airgradienthq/airgradient-client) and [airgradient-ota](https://github.com/airgradienthq/airgradient-ota) as a submodule
8. Update howto-compile docs for this changes to include submodule

#### Last time fetch configuration or post measures to airgradient server failed

WhatsApp Image 2025-03-27 at 3 58 34 PM (1)

#### Network not connected (wifi or cellular)

WhatsApp Image 2025-03-27 at 4 02 10 PM

#### Offline mode

WhatsApp Image 2025-03-27 at 4 26 11 PM


examples/OneOpenAir/OpenMetrics.cpp
```diff
@@ -43,13 +45,13 @@ String OpenMetrics::getPayload(void) {
...
              "gauge");
-  add_metric_point("", apiClient.isFetchConfigurationFailed() ? "0" : "1");
+  add_metric_point("", agClient->isLastFetchConfigSucceed() ? "1" : "0");
 
   add_metric(
       "post_ok",
       "1 if the AirGradient device was able to successfully send to the server",
       "gauge");
-  add_metric_point("", apiClient.isPostToServerFailed() ? "0" : "1");
+  add_metric_point("", agClient->isLastPostMeasureSucceed() ? "1" : "0");
```
nick-4711, line 54: hasLastMeasurePostSuceeded

src/AgOledDisplay.cpp
```diff
@@ -5,12 +5,31 @@
 /** Cast U8G2 */
 #define DISP() ((U8G2_SH1106_128X64_NONAME_F_HW_I2C *)(this->u8g2))
 
+static const unsigned char wifi_issue_bits[] = {
```
nick-4711, line 8: I'm uncertain if we need to capitalize static consts?

src/AgValue.cpp
```diff
@@ -686,6 +688,153 @@ float Measurements::getCorrectedPM25(bool useAvg, int ch, bool forceCorrection)
   return corrected;
 }
 
+Measurements::MeasurementCycle Measurements::getMeasurementCycle() {
+  MeasurementCycle mc;
```
nick-4711, line 692: I think this not the "cycle" it is more what we have measured during a cycle?

examples/OneOpenAir/OpenMetrics.cpp
samuelbles07: I think both is right. But to follow convention, would be better to have a specific PR for it.

examples/OneOpenAir/OpenMetrics.cpp
```diff
@@ -43,13 +45,13 @@ String OpenMetrics::getPayload(void) {
              "1 if the AirGradient device was able to successfully fetch its "
              "configuration from the server",
              "gauge");
-  add_metric_point("", apiClient.isFetchConfigurationFailed() ? "0" : "1");
+  add_metric_point("", agClient->isLastFetchConfigSucceed() ? "1" : "0");
```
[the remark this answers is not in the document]
samuelbles07: I think both is right. But to follow convention, would be better to have a specific PR for it.

src/AgValue.cpp
samuelbles07: Update

src/AgOledDisplay.cpp
samuelbles07: Fixed

PR #302, opened 2025-04-09
Improve cellular client reconnection

samuelbles07:
## Changes

1. Power cycle cellular module control directly on main, if airgradient client not ready for more than 1 hour
2. If in 2 hour airgradient client still not ready, then esp restart. Has redundant check in both main and networking tasks.
3. Transmit measures only if measurement cycle queue size is divisible by 3

### DIFF
--- a/examples/OneOpenAir/OneOpenAir.ino
+++ b/examples/OneOpenAir/OneOpenAir.ino
@@ -120,6 +120,9 @@ static bool ledBarButtonTest = false;
 static String fwNewVersion;
 static int lastCellSignalQuality = 99; // CSQ 
 
+// Default value is 0, indicate its not started yet
+uint32_t startTimeClientNotReady = 0;
+
 SemaphoreHandle_t mutexMeasurementCycleQueue;
 static std::vector<Measurements::Measures> measurementCycleQueue;
 
@@ -147,6 +150,7 @@ static void displayExecuteOta(AirgradientOTA::OtaResult result, String msg, int 
 static int calculateMaxPeriod(int updateInterval);
 static void setMeasurementMaxPeriod();
 static void newMeasurementCycle();
+static void checkCellularClientNotReady(); 
 static void networkSignalCheck();
 static void networkingTask(void *args);
 
@@ -299,6 +303,12 @@ void setup() {
 }
 
 void loop() {
+  if (networkOption == UseCellular) {
+    // Check if cellular client not ready until certain time
+    // Redundant check in both task to make sure its executed 
+    checkCellularClientNotReady();
+  }
+
   // Schedule to feed external watchdog
   watchdogFeedSchedule.run();
 
@@ -1494,6 +1504,30 @@ void networkSignalCheck() {
   }
 }
 
+/**
+* If in 2 hours still not ready, then restart esp
+*/
+void checkCellularClientNotReady() {
+  if (startTimeClientNotReady > 0 &&
+      (millis() - startTimeClientNotReady) > (2 * 60 * 60000)) {
+    // Give up wait
+    Serial.println("CLIENT NOT READY TAKING TOO LONG! Rebooting ...");
+    int i = 3;
+    while (i != 0) {
+      if (ag->isOne()) {
+        String tmp = "Reboot in " + String(i);
+        oledDisplay.setText("CE error", "too long", tmp.c_str());
+      } else {
+        Serial.println("Rebooting... " + String(i));
+      }
+      i = i - 1;
+      delay(1000);
+    }
+    oledDisplay.setBrightness(0);
+    esp_restart();
+  }
+}
+
 void networkingTask(void *args) {
   // OTA check on boot
 #ifdef ESP8266
@@ -1514,9 +1548,6 @@ void networkingTask(void *args) {
     sendDataToServer();
     measurementSchedule.update();
   }
-
-  // Default value is 0, indicate its not started yet
-  uint32_t startTimeClientNotReady = 0;
 
   // Reset scheduler
   configSchedule.update();
@@ -1543,24 +1574,10 @@ void networkingTask(void *args) {
         // Enable at command debug
         agSerial->setDebug(true); 
 
-        // If in 2 hours still not ready, then restart esp
-        if ((millis() - startTimeClientNotReady) > (2 * 60 * 60000)) {
-          Serial.println("CLIENT NOT READY TAKING TOO LONG! Rebooting ...");
-          int i = 3;
-          while (i != 0) {
-            if (ag->isOne()) {
-              String tmp = "Reboot in " + String(i);
-              oledDisplay.setText("CE error", "too long", tmp.c_str());
-            } else {
-              Serial.println("Rebooting... " + String(i));
-            }
-            i = i - 1;
-            delay(1000);
-          }
-          oledDisplay.setBrightness(0);
-          esp_restart();
-        }
-        
+        // Check if cellular client not ready until certain time
+        // Redundant check in both task to make sure its executed 
+        checkCellularClientNotReady();
+
         // Starting from 1 hour since its not ready, power cycling the module
         bool resetModule = true;
         if ((millis() - startTimeClientNotReady) > (60 * 60000)) {
--- a/examples/OneOpenAir/OneOpenAir.ino
+++ b/examples/OneOpenAir/OneOpenAir.ino
@@ -1571,9 +1571,6 @@ void networkingTask(void *args) {
   configSchedule.update();
   transmissionSchedule.update();
 
-
-  uint32_t startTime = millis();
-
   while (1) {
     // Handle reconnection based on mode
     if (networkOption == UseWifi) {
--- a/examples/OneOpenAir/OneOpenAir.ino
+++ b/examples/OneOpenAir/OneOpenAir.ino
@@ -37,6 +37,7 @@ CC BY-SA 4.0 Attribution-ShareAlike 4.0 International License
 #include "Arduino.h"
 #include "EEPROM.h"
 #include "ESPmDNS.h"
+#include "Libraries/airgradient-client/src/common.h"
 #include "LocalServer.h"
 #include "MqttClient.h"
 #include "OpenMetrics.h"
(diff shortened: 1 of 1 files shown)


examples/OneOpenAir/OneOpenAir.ino
```diff
@@ -1529,14 +1576,42 @@ void networkingTask(void *args) {
     }
     else if (networkOption == UseCellular) {
       if (agClient->isClientReady() == false) {
-        Serial.println("Cellular client not ready, ensuring connection...");
+        // Start time if value still default
+        if (startTimeClientNotReady == 0) {
+          startTimeClientNotReady = millis();
```
nick-4711, line 1581: The current name suggests that it is a boolean (but it is a timestamp). Propose to rename to 

agCeClientProblemDetectedTimestamp

or similar

examples/OneOpenAir/OneOpenAir.ino
```diff
@@ -1493,6 +1516,30 @@ void networkSignalCheck() {
...
 }
 
+/**
+* If in 2 hours still not ready, then restart esp
+*/
+void checkCellularClientNotReady() {
+  if (startTimeClientNotReady > 0 &&
+      (millis() - startTimeClientNotReady) > (2 * 60 * 60000)) {
+    // Give up wait
+    Serial.println("CLIENT NOT READY TAKING TOO LONG! Rebooting ...");
```
nick-4711, line 1526: Propose to log the time (2h), maybe something like

Rebooting because CE client issues for 2h detected.

examples/OneOpenAir/OneOpenAir.ino
```diff
@@ -1493,6 +1516,30 @@ void networkSignalCheck() {
...
+    int i = 3;
+    while (i != 0) {
+      if (ag->isOne()) {
+        String tmp = "Reboot in " + String(i);
```
nick-4711, line 1530: Rebooting in ..

examples/OneOpenAir/OneOpenAir.ino
```diff
@@ -1493,6 +1516,30 @@ void networkSignalCheck() {
...
+        oledDisplay.setText("CE error", "too long", tmp.c_str());
```
nick-4711, line 1531: CE communication error
or
Cellular communication error
if we have enough space, probably not

examples/OneOpenAir/OneOpenAir.ino
```diff
@@ -1529,14 +1576,42 @@ void networkingTask(void *args) {
     }
     else if (networkOption == UseCellular) {
       if (agClient->isClientReady() == false) {
-        Serial.println("Cellular client not ready, ensuring connection...");
+        // Start time if value still default
+        if (startTimeClientNotReady == 0) {
+          startTimeClientNotReady = millis();
+        }
+
+        // TODO: Need to handle if millis is overflow (back to 0)
```
nick-4711, line 1584: Propose to keep seconds in something larger than uint32_t

examples/OneOpenAir/OneOpenAir.ino
```diff
@@ -1529,14 +1576,42 @@ void networkingTask(void *args) {
...
         agSerial->setDebug(true); 
-        if (agClient->ensureClientConnection() == false) {
-          Serial.println("Cellular client connection not ready, retry in 5s...");
-          delay(5000);
+
+        // Check if cellular client not ready until certain time
+        // Redundant check in both task to make sure its executed 
+        checkCellularClientNotReady();
+
+        // Starting from 1 hour since its not ready, power cycling the module
```
nick-4711, line 1593: I don't understand this sentence

examples/OneOpenAir/OneOpenAir.ino
```diff
@@ -1529,14 +1576,42 @@ void networkingTask(void *args) {
...
+        bool resetModule = true;
+        if ((millis() - startTimeClientNotReady) > (60 * 60000)) {
+          Serial.println("Power cycling module");
```
nick-4711, line 1596: Propose to also say why and what

Power cycling CE-card due to network issues for more than 1h

or similar

examples/OneOpenAir/OneOpenAir.ino
```diff
@@ -1529,14 +1576,42 @@ void networkingTask(void *args) {
...
+          cell->powerOff();
```
nick-4711, line 1597: cell -> cellularCard

examples/OneOpenAir/OneOpenAir.ino
```diff
@@ -1529,14 +1576,42 @@ void networkingTask(void *args) {
...
+          delay(2000);
+          cell->powerOn();
+          delay(10000);
+          resetModule = false; // No need to reset module anymore
```
nick-4711, line 1601: I don't understand the comment, we might reset it again in the future?

nick-4711:
In the lib

"Failed reinitialized cellular module
->
"Failed to reinitialize the cellular module

nick-4711:
Propose to also add a sample log where we see the two mechanisms kick in to the PR

examples/OneOpenAir/OneOpenAir.ino
samuelbles07: Already tried both, not enough space

examples/OneOpenAir/OneOpenAir.ino
samuelbles07: So basically just to make sure when CE client has a problem, the check happen in main and networking task, just in case one of the task somehow stuck.

examples/OneOpenAir/OneOpenAir.ino
samuelbles07: It is a local variable, that will be passed to `ensureClientConnection()` . Already modified the comment.

examples/OneOpenAir/OneOpenAir.ino
nick-4711: I mean I don't understand "Starting from 1 hour since its not ready"

Maybe we should remove the commend and make the log output more clear

if ((millis() - startTimeClientNotReady) > (60 * 60000)) {

          Serial.println("The CE client hasn't recovered in xh, performing a power cycle"); 

if we assume that the wording will be something like

examples/OneOpenAir/OneOpenAir.ino
nick-4711: For me the module is the code and the card the hardware. Here we have a problem with the hardware (or network), not with the code? Or does cellularModule refer to the A7672G chip on the CE card? Then there could still be issues like e.g. the A7672G not being able to communicate with the SIM card, so cellularModule referring to the A7672G would be misleading in such a case?

examples/OneOpenAir/OneOpenAir.ino
samuelbles07: Yes, that's what i though too. Implemented.

examples/OneOpenAir/OneOpenAir.ino
samuelbles07: Ahh, we have a different definition then. I guess the convention in c or c++ is never saying module, instead of library. But even though if it is, we never add module when declaring variable or library (eg. agClient not agClientLibrary).

> Or does cellularModule refer to the A7672G chip on the CE card

yes

I'm not agree with CE card because the chip is not actually communicate with the sim card, but the module.  I think we can say sim card as the part of the whole module, since its actually can be a common sim card or e-sim. What you think?

samuelbles07:
> Propose to also add a sample log where we see the two mechanisms kick in to the PR

Attached the logs, module unplugged around _17:28_

Point of interest: 

1. _16:28:28_ start detect client not ready
2. _17:28:28_ start power cycle module in 1 hour mark
3. _18:26:53_ system restart

[2hours.log](https://github.com/user-attachments/files/19684237/2hours.log)

examples/OneOpenAir/OneOpenAir.ino
nick-4711: OK, yes, library for the software makes sense. But the hardware we called CE-Card from the very start, not CE-module.

examples/OneOpenAir/OneOpenAir.ino
```diff
@@ -1493,6 +1521,31 @@ void networkSignalCheck() {
   }
 }
 
+/**
+* If in 2 hours still not ready, then restart esp
+*/
+void checkCellularClientNotReady() {
```
nick-4711, line 1527: I still don't like the name, we can't guess what it does.

examples/OneOpenAir/OneOpenAir.ino
```diff
@@ -1493,6 +1521,31 @@ void networkSignalCheck() {
...
+  if (agCeClientProblemDetectedTime > 0 &&
+      (MICROS_TO_MINUTES() - agCeClientProblemDetectedTime) >
+          TIMEOUT_WAIT_FOR_CELLULAR_MODULE_READY) {
+    // Give up wait
+    Serial.println("Rebooting because CE client issues for 2 hours detected");
+    int i = 3;
+    while (i != 0) {
+      if (ag->isOne()) {
+        String tmp = "Rebooting in " + String(i);
+        oledDisplay.setText("CE error", "since 1h", tmp.c_str());
```
nick-4711, line 1537: should it say 2h

examples/OneOpenAir/OneOpenAir.ino
samuelbles07: Updated to `restartIfCeClientIssueOverTwoHours`

examples/OneOpenAir/OneOpenAir.ino
```diff
@@ -1529,14 +1582,43 @@ void networkingTask(void *args) {
     }
     else if (networkOption == UseCellular) {
       if (agClient->isClientReady() == false) {
-        Serial.println("Cellular client not ready, ensuring connection...");
+        // Start time if value still default
+        if (agCeClientProblemDetectedTime == 0) {
+          agCeClientProblemDetectedTime = MICROS_TO_MINUTES();
```
nick-4711, line 1587: minor, I think it should be named
minutes() 
or so (similar to millis()), we do not pass anything in

nick-4711 (approved):

PR #303, opened 2025-04-13
Fix incorrect TVOC / NOx values when when network option is cellular 

samuelbles07:
## Changes

- Add a pause and resume SGP41 reading to its task, so no need to reset SGP41 algorithm object
- Use latest airgradient-ota library, by improving ota sequence to have a more accurate event. Please see: #https://github.com/airgradienthq/airgradient-ota/pull/2 . Because of this, no need to pause SGP41 task every time doing OTA check. Only pause when actually know that firmware update is available.

### DIFF
--- a/examples/OneOpenAir/OneOpenAir.ino
+++ b/examples/OneOpenAir/OneOpenAir.ino
@@ -584,7 +584,7 @@ void otaHandlerCallback(AirgradientOTA::OtaResult result, const char *msg) {
     Serial.println("Firmware update starting...");
     if (configuration.hasSensorSGP && networkOption == UseCellular) {
       // Temporary pause SGP41 task while cellular firmware update is in progress
-      ag->sgp41.pauseHandle();
+      ag->sgp41.pause();
     }
     displayExecuteOta(result, fwNewVersion, 0);
     break;
@@ -594,13 +594,15 @@ void otaHandlerCallback(AirgradientOTA::OtaResult result, const char *msg) {
     displayExecuteOta(result, "", std::stoi(msg));
     break;
   case AirgradientOTA::Failed:
+      displayExecuteOta(result, "", 0);
+      if (configuration.hasSensorSGP && networkOption == UseCellular) {
+        // Cellular firmware update finish, resuming SGP41 task
+        ag->sgp41.resume();
+      }
+      break;
   case AirgradientOTA::Skipped:
   case AirgradientOTA::AlreadyUpToDate:
     displayExecuteOta(result, "", 0);
-    if (configuration.hasSensorSGP && networkOption == UseCellular) {
-      // Cellular firmware update finish, resuming SGP41 task
-      ag->sgp41.resumeHandle();
-    }
     break;
   case AirgradientOTA::Success:
     displayExecuteOta(result, "", 0);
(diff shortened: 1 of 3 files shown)


examples/OneOpenAir/OneOpenAir.ino
```diff
@@ -602,6 +597,10 @@ void otaHandlerCallback(AirgradientOTA::OtaResult result, const char *msg) {
   case AirgradientOTA::Skipped:
   case AirgradientOTA::AlreadyUpToDate:
     displayExecuteOta(result, "", 0);
+    if (configuration.hasSensorSGP && networkOption == UseCellular) {
```
nick-4711, line 600: I thought the idea was to not interrupt the polling if no firmware update is needed. So I'm a bit surprised that we need to resume here.

src/Sgp41/Sgp41.cpp
```diff
@@ -131,6 +131,22 @@ void Sgp41::handle(void) {
 }
 
 #else
+
+void Sgp41::pauseHandle() {
```
nick-4711, line 135: propose to rename to simply "pause()" and "resume()"

src/Sgp41/Sgp41.cpp
```diff
@@ -131,6 +131,22 @@ void Sgp41::handle(void) {
...
+  // Set latest value to invalid
+  tvocRaw = utils::getInvalidVOC();
+  tvoc = utils::getInvalidVOC();
+  noxRaw = utils::getInvalidNOx();
+  nox = utils::getInvalidNOx();
+} 
+
+void Sgp41::resumeHandle() {
+  onPause = false;
+  Serial.println("Resume SGP41 handler task");
```
nick-4711, line 147: Since we used the -ing form in pause(), I propose to also use it here

-> Resuming SGP41 ...

nick-4711:
Propose to say "incorrect values" in the title of the PR

-> Fix incorrect TVOC / NOx values when ...

examples/OneOpenAir/OneOpenAir.ino
samuelbles07: Yes, you're right, even if it has no impact because it haven't paused yet. Fixed it.

https://github.com/airgradienthq/arduino/pull/303/commits/e7603a76590bd0994e2cb24b6e7aba9113565013#diff-a9eec5b0808e2ecb2731d6bcffbc2fd06fefc1b8643ff524865db674caf59fb5R597

examples/OneOpenAir/OneOpenAir.ino
```diff
@@ -594,13 +594,15 @@ void otaHandlerCallback(AirgradientOTA::OtaResult result, const char *msg) {
     displayExecuteOta(result, "", std::stoi(msg));
     break;
   case AirgradientOTA::Failed:
+      displayExecuteOta(result, "", 0);
+      if (configuration.hasSensorSGP && networkOption == UseCellular) {
+        // Cellular firmware update finish, resuming SGP41 task
```
nick-4711, line 599: I think it failed (and hasn't finished). We don't need the comment I would say.

nick-4711 (approved):

Issue #309, opened 2025-04-30
I-9PSL unit reset itself when calling /metrics endpoint


m4r1k:
Hey there,

When curling the `/metrics` endpoint, much to my surprise, the AirGradient ONE reset itself.

AirGradient ONE DIY (I-9PSL)
Firmware: 3.3.6

config
```json
{
  "country": "IT",
  "pmStandard": "ugm3",
  "ledBarMode": "co2",
  "abcDays": 8,
  "tvocLearningOffset": 12,
  "noxLearningOffset": 12,
  "mqttBrokerUrl": "",
  "temperatureUnit": "c",
  "configurationControl": "local",
  "postDataToAirGradient": true,
  "ledBarBrightness": 100,
  "displayBrightness": 100,
  "offlineMode": false,
  "monitorDisplayCompensatedValues": false,
  "model": "I-9PSL",
  "httpDomain": "",
  "disableCloudConnection": false
}
```

is this the expected behavior? See the video below.

https://github.com/user-attachments/assets/6d010791-2c86-49f7-936a-9bba38a753d1

ccoley:
I can confirm this issue as well. I have 4 monitors on firmware v3.3.6 and a prometheus instance pulling metrics from the `/metrics` endpoint regularly. All of the monitors have been in a reset loop for about a day, maybe longer.

I downgraded one the monitors to v3.1.13 manually, and that resolved the issue.
I then upgraded that monitor to v3.2.0 and the issue is still resolved.
If I stop prometheus from polling the monitors, that also resolves the issue.

ccoley:
I see this error in the console when the reset happens on v3.3.6:

```
Guru Meditation Error: Core  0 panic'ed (Load access fault). Exception was unhandled.

Core  0 register dump:
MEPC    : 0x4202d16e  RA      : 0x42004aa2  SP      : 0x3fcb34f0  GP      : 0x3fc8e800
TP      : 0x3fc6e56c  T0      : 0x61204550  T1      : 0x6b6f5f67  T2      : 0x59542023
S0/FP   : 0x3fc920dc  S1      : 0x3c14f000  A0      : 0x00000000  A1      : 0x00000000
A2      : 0x00000001  A3      : 0x00000000  A4      : 0x00000000  A5      : 0x00000080
A6      : 0x000a6567  A7      : 0x75616720  S2      : 0x3c159000  S3      : 0x3c133000
[...]

PR #310, opened 2025-05-01
Fix calling airgradient client before initialization on open metrics

samuelbles07:
## Changes

Previously, `airgradientClient` was passed to `openmetrics` before it was initialized. This has been fixed by separating the passing of `airgradient` and `airgradientClient` to `openmetrics`, allowing `airgradientClient` to be passed only after it has been initialized.


ccoley:
This resolves #309

PR #362, opened 2026-02-23
PM2.5 correction by PM2.5 Raw

samuelbles07:
## Changes

1. PM2.5 correction by PM2.5 Raw
2. Fix PM2.5 correction by 0.3 Particle Count by add dedicated function
3. Fix batch correction

### DIFF
--- a/src/AgConfigure.cpp
+++ b/src/AgConfigure.cpp
@@ -130,9 +130,9 @@ PMCorrectionAlgorithm Configuration::matchPmAlgorithm(String algorithm) {
   // If string not match from enum, check if correctionAlgorithm is one of the PM batch corrections
   if (result == COR_ALGO_PM_UNKNOWN) {
     // Check the substring "slr_PMS5003_xxxxxxxx"
-    if (algorithm.substring(0, 11) == "slr_PMS5003") {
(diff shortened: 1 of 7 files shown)


petercrona:
@samuelbles07 I tried this out on my device, and no longer see 0 all the time :)  Thanks for fixing this!

Not directly issues, but some observations, and perhaps future work to consider:

* PM1 and PM10 are still 0 - looks a little strange, especially as PM2.5 is a subset of PM10, so arguably `PM10_compensated = MAX(PM_2_5_COMPENSATED, PM10_RAW)` would already be a small win.
* PM2.5 on the device's screen and in app.airgradient.com deviates - compensation is only applied on device and not used by dashboard?

Also, perhaps you know, but firmware size is approaching the limit. I played around with persisting VOC Index state (https://sensirion.github.io/gas-index-algorithm/api.html#sensirion_gas_index_algorithm.voc_algorithm.VocAlgorithm.set_states and https://sensirion.github.io/gas-index-algorithm/api.html#sensirion_gas_index_algorithm.voc_algorithm.VocAlgorithm.get_states ), and accidentally put my device in a bootloop (it stopped starting) by flashing with a too large firmware :facepalm:. Managed to fix by flashing a smaller firmware.

PR #363, opened 2026-03-08
fix: include factory_calibration in PM correction algorithm matching

BkSouX:
Description:                                                                                                                  
  `matchPmAlgorithm()` loop bound was hardcoded to `COR_ALGO_PM_SLR_CUSTOM + 1`,
  so it never checked index 4 (`COR_ALGO_PM_SLR_FACTORY_CALIBRATION`).                                                          
                                                                   
  Bug introduced in b374a9a (Jan 2026) — the commit added the enum value and the
  name table entry but missed updating the loop bound.

  **Impact:** devices configured with factory calibration silently got uncorrected
  PM2.5 readings.

### DIFF
--- a/src/AgConfigure.cpp
+++ b/src/AgConfigure.cpp
@@ -116,7 +116,7 @@ PMCorrectionAlgorithm Configuration::matchPmAlgorithm(String algorithm) {
   // If the input string matches an algorithm name, return the corresponding enum value
   // Else return Unknown
 
-  const size_t enumSize = COR_ALGO_PM_SLR_FACTORY_CALIBRATION + 1;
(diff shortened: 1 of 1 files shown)


samuelbles07:
Hi @BkSouX ,

Thanks for noticing this! Anyway, we already add another algorithm type on this PR #362 , so maybe you can adjust it?

Check this [line](https://github.com/airgradienthq/arduino/pull/362/changes#diff-14434cc682e7a4d4df8b8eb28077af67d1990888a47ad06207c1b99dc69b3545R30).

We plan to release new version next week. So this will be included.

BkSouX:
I've updated the fix to derive the enum size from the names array using sizeof, so it won't break again when new algorithms are added.

samuelbles07:
Even better, thanks @BkSouX !

Issue #387, opened 2026-07-24
Github Actions are logging warnings about deprecated Node20


MallocArray:
At the end of Actions runs, a warning is being logged

Warning: Node.js 20 is deprecated. The following actions target Node.js 20 but are being forced to run on Node.js 24: actions/checkout@v4.2.2, actions/setup-python@v5. For more information see: https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/

Update the 2 actions to later versions that address this deprecation

PR incoming

PR #388, opened 2026-07-24
Update actions that use Nod20 which is EOL

MallocArray:
https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/

Addresses #387 

Warning: Node.js 20 is deprecated. The following actions target Node.js 20 but are being forced to run on Node.js 24: actions/checkout@v4.2.2, actions/setup-python@v5. For more information see: https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -3,7 +3,7 @@ jobs:
   trailing-whitespace:
     runs-on: ubuntu-24.04
     steps:
-      - uses: actions/checkout@v4.2.2
+      - uses: actions/checkout@v7.0.1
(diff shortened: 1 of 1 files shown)


MallocArray:
Or most examples have instead of `uses: actions/checkout@v7.0.1` they just have `uses: actions/checkout@v7` so any minor updates to the 7.x line are used.  Happy to change to lock to the major version instead of pinning to a particular release, as updates do come out regularly and if not maintained, these will get behind.

Issue #386, opened 2026-07-24
ONE firmware boot-loops when following howto-compile.md instructions, with suggested resolution


seansfkelley:
Following https://github.com/airgradienthq/arduino/blob/e5a3ddccc408869a2c5fb21c5d9e65f2ed55e90b/docs/howto-compile.md, the logs for my ONE would boot loop several times a second with this repeated ad nauseam:

```
ESP-ROM:esp32c3-api1-20210207
Build:Feb  7 2021
rst:0x3 (RTC_SW_SYS_RST),boot:0xc (SPI_FAST_FLASH_BOOT)
Saved PC:0x403cf94c
SPIWP:0xee
mode:DIO, clock div:1
load:0x3fcd5810,len:0x438
load:0x403cc710,len:0x90c
load:0x403ce710,len:0x2624
entry 0x403cc710
```

I sent Claude on a bender and it decided the issue was that the binary produced by these instructions is too large for the ROM. I did _not_ confirm this as the direct cause because in so doing Claude pointed out that the production builds use PlatformIO, so I opted to go straight for using PlatformIO and bypass the Arduino IDE completely. **This does fix the issue**, though I have not confirmed which exact change did it.

To build with PlatformIO locally, I:

1. follow the existing instructions to pull all the necessary libraries
2. `brew install platformio` (this is on a Mac)
3. peg `platform = espressif32` to `platform = espressif32@7.0.1` in `platformio.ini` (just to double make sure I get 2.0.17 per the instructions -- this version is indeed based on 2.0.17)
4. `ln -s AirGradient_Air_Quality_Sensor arduino` in the Arduino libraries directory (more on this later)
5. build and flash from the command line
    ```sh
    pio run -e esp32-c3
    pio run -e esp32-c3 -t erase
    pio run -e esp32-c3 -t upload --upload-port /dev/cu.usbmodem101 # or whatever your port is called
    ```
6. watch the logs with `pio device monitor -e esp32-c3 -p /dev/cu.usbmodem101`

I haven't opened a PR because there are a number of decisions that are stylistic or editorial:

- I'm not sure what level you want the instructions to be at, but I could see that straight CLI commands (instead of using the Arduino IDE) might be too unfriendly for beginners.
- I haven't figured out how much of this still requires the Arduino IDE for setup.
[...]

PR #392, opened 2026-07-27
[Experimental] Add HomeKit support based on HomeSpan for the ONE.

seansfkelley:
## Summary and Features

This PR adds support for reporting to HomeKit using [HomeSpan](https://github.com/HomeSpan/HomeSpan) for the ONE. (It is the only one I have and so the only one I tested.)

When set up, the ONE reports all the readings it has, with some caveats (see below). It appears in HomeKit as little pills, since it's a read-only sensor and not an interactive thing.

Screenshot 2026-07-27 at 10 29 57

To try it out, after flashing this firmware, pair with the code **466-37-726** (HomeSpan's default value).

## Technical Details and Caveats

This PR has significant shortcomings, however, so it needs input from other HomeKit users (i.e. what they want) and AirGradient (i.e. what is acceptable) if it is to be incorporated upstream. That said, I'm opening this PR for discoverability for other potential HomeKit tinkerers.

The major caveats are:

- HomeSpan adds ~200KB, which pushes the binary size over the 1.9MB limit. (See also #386.) This PR switches to the "Huge APP" configuration, which **disables OTA updates** due to lack of space.
- HomeKit wants absolute ug/m3 measurements for VOC and NOx, but the ONE only reports relative indices. This PR just **reports the index as if it's the concentration**. (Also HomeKit uses the terminology "NO2", not the broader "NOx".)

And these aren't really caveats so much as things to be aware of/corners I cut temporarily:

- The air quality index is currently reported as the max of the CO2/PM2.5 index because I didn't want to go down the rabbit hole of computing the full formulas.
- HomeSpan is pegged to 1.9.1 because that's the last version that supports this board.
- I had to configure HomeSpan with a non-default port (here, 1201) because otherwise it would fight with the API server on 80.
- HomeKit hides the specific readings (like PM2.5) behind a tap, which is annoying if you want the details up-front, and I can't figure out how to configure it otherwise. For some reason it also lists the CO2 sensor as just "Carbon Dioxide Sensor" without giving you figures until you tap into it. Perhaps third-party HomeKit viewers (Eve?) don't have this problem.

### DIFF
--- a/examples/OneOpenAir/OneOpenAir.ino
+++ b/examples/OneOpenAir/OneOpenAir.ino
@@ -63,6 +63,7 @@ CC BY-SA 4.0 Attribution-ShareAlike 4.0 International License
 #include "Libraries/airgradient-ota/src/airgradientOtaCellular.h"
 #include "esp_system.h"
 #include "freertos/projdefs.h"
+#include "homekit.h"
 
 #define LED_BAR_ANIMATION_PERIOD 100                       /** ms */
 #define DISP_UPDATE_INTERVAL 2500                          /** ms */
@@ -165,6 +166,8 @@ static AirgradientClient::PayloadType getClientPayloadType();
 static AirgradientClient::CommonPayload buildCommonPayload(Measurements::Measures &mc);
 static void saveOperatorState();
 static void restoreOperatorState();
+static void setupHomeKit();
+static void updateHomeKit();
 
 AgSchedule dispLedSchedule(DISP_UPDATE_INTERVAL, updateDisplayAndLedBar);
 AgSchedule configSchedule(WIFI_SERVER_CONFIG_SYNC_INTERVAL, configurationUpdateSchedule);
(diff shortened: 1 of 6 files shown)


examples/OneOpenAir/OneOpenAir.ino
```diff
@@ -178,6 +181,11 @@ AgSchedule watchdogFeedSchedule(60000, wdgFeedUpdate);
 AgSchedule checkForUpdateSchedule(FIRMWARE_CHECK_FOR_UPDATE_MS, checkForFirmwareUpdate);
 AgSchedule networkSignalCheckSchedule(10000, networkSignalCheck);
 AgSchedule printMeasurementsSchedule(6000, printMeasurements);
+AgSchedule homekitSchedule(10000, updateHomeKit);
+
+// HomeKit pairing (SRP) is stack-heavy; enlarge the Arduino loop-task stack so
+// homeSpan.poll() has room to complete pairing without a stack overflow.
+SET_LOOP_TASK_STACK_SIZE(16 * 1024);
```
seansfkelley, line 188: I'm not sure if this is actually necessary. Would check it again before committing to it.

PR #393, opened 2026-07-28
fix(OneOpenAir): make board type detection robust to OLED failure

java-devil:
## Problem
The unified firmware decides the board type on every boot by probing the I2C bus for the OLED display. If a ONE's OLED display fails to properly `ACK` (either transiently or persistently) - an otherwise perfectly functional device is being misidentified as an Open Air, which results in constant garbage temperature/humidity (**−6.7 °C / 7.3 %**) measurements.

## Proposed Fix
- identify the board as `ONE_INDOOR` if **either** the OLED (`0x3C`) **or** the SHT4x/SHT3x (`0x44`/`0x45`) `ACK`
- identify the board as `OPEN_AIR_OUTDOOR` if x3 retries (~200 ms apart) fail to `ACK` (a real Open Air pays ~400ms)


samuelbles07:
Hi @java-devil , we decide to use other approach, you can see #394

